PR #2, opened 2018-08-28
Typo: addField instead of AddFields in crud-tutorial.md

ikkebra:
When the tutorial is telling how to add the model's 'name' field to the controller, it is using addFields instead of addField   ('s' needs removed)


tabacitu:
True! Thank you @ikkebra !

PR #4, opened 2018-08-30
save uploaded image as jpg, don't just rename

markcollister:
This might save some people time, as per usual I just copied the documentation and the example was saving / renaming my png's to jpg's


tabacitu:
That's a good update - thank you @markcollister !

PR #5, opened 2018-09-06
Added missing / causing broken links

Patabugen:
The links under "Additionally, here are a few more ways you can customize your CRUDs:" were missing their forward slashes at the start, causing the links to land at 404s.


tabacitu:
True. Thank you @Patabugen !

PR #6, opened 2018-09-20
Update crud-columns.md

jackfruh:
adjusted wording in a callout.
added a callout in the responsive section 'calling out' that the elipsis that sometimes shows up is part of the responsive design and is not a user addressable property.


tabacitu:
All true. Thank you @jackfruh !

PR #9, opened 2018-10-29
Reorder : integer fields

MarcBrillault:
As answered in this StackOverflow question : https://stackoverflow.com/questions/40334901/what-are-the-reordering-columns-on-laravel-backpack/40361674#40361674


tabacitu:
Totally true. Thank you @MarcBrillault !

PR #10, opened 2018-10-30
Moving the "readonly" and "disabled" into the attributes

seandowney:
"readonly" doesn't work outside the "attributes"


tabacitu:
Totally! Thank you @seandowney !

PR #13, opened 2018-11-15
Add Mandatory Primary Key topic

edgareler:
As the method [updateCrud()](https://github.com/Laravel-Backpack/CRUD/blob/master/src/app/Http/Controllers/Operations/Update.php#L60) requires the primary key name as a field to update the row, it's important to make clear in the documentation that the primary key is a mandatory field in the CRUD setup.

### DIFF
--- a/3.4/crud-fields.md
+++ b/3.4/crud-fields.md
@@ -90,6 +90,16 @@ $this->crud->addField($field_definition_array)->beforeField('name');
 $this->crud->addField($field_definition_array)->afterField('name');
 ```
 
+<a name="mandatory-primary-key"></a>
+### Mandatory Primary Key
+
+To enable the update actions in your CRUD you must add the primary key as one of its fields. The primary key is used to find the row that will be updated.
+
+If you don't add the primary key as a field of your CRUD, Backpack will return the following error when you try to update a record: `404 Page not found. No query results for model App\YourModel`.
+
+If you don't want to display the primary key in your form, set its field type to [hidden](/docs/{{version}}/crud-fields#hidden).
+
+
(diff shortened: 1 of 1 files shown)


tabacitu:
Hi @edgareler ,

Thanks for the docs update - but that’s actually not true. You _don’t_ need to add the primary key as a field, [Backpack will add that one for you](https://github.com/Laravel-Backpack/CRUD/blob/master/src/PanelTraits/Update.php#L66-L73). But if you _do_ add one, Backpack will use yours instead of the standard “hidden” one it adds.

The error you got must be from an ill-defined primary key on the model or something… Try adding an item through the Bp interface, then editing the same item - it should work, even without a field for the primary key.

Cheers!

tabacitu:
Thanks for the docs update though - not everybody want to help others after they struggle with something. Really appreciate it!

edgareler:
Hi @tabacitu, thanks for the response.

After analyzing this `getUpdateFields()` method, I understood what happened.

I created a custom field type, which I called "static", to just display a certain field name and value inside a `<div>`, without adding an `<input type="hidden">` for it. When I added my `id` field as `static` to my CRUD setup, that replaced the "id" field you added automatically on the `getUpdateFields()`. Then, when I submitted the update form, I got the `404` error, because the primary key field didn't exist in the `$data` array (as it was not a form field).

Given this fact, I understand that this PR I sent is no more appropriate. But you should add this information about the automatic ID field somewhere in the documentation, otherwise more devs can struggle with this same issue like me. I just found the reason of this issue (missing key) after a deep debug inside Backpack code.

PR #14, opened 2018-11-19
Details on reorder fields

MarcBrillault:
I forgot to update the default values and nullable field last time.


tabacitu:
Thank you!

PR #22, opened 2018-11-23
Correct method names for clone operation

mattsches:


tabacitu:
Totally! Thank you @mattsches !

PR #23, opened 2018-11-28
removed unused parameter from 3.5

urlportal:
## Overview

**File**: src / PanelTraits / Fields.php

before v3.4.39 had:  
`checkIfFieldIsFirstOfItsType( $field, $fields_array )`

v3.4.39 and later (inc. v3.5):  
`checkIfFieldIsFirstOfItsType( $field )`

**Result**: Removed second parameter from 3.5-docs example

### DIFF
--- a/3.5/crud-fields.md
+++ b/3.5/crud-fields.md
@@ -1637,7 +1637,7 @@ And your blade file something like:
 </div>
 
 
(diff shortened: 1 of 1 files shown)


tabacitu:
True, thank you!

PR #24, opened 2018-11-28
proposed composer install instead of composer update

MarcBrillault:
I'm strongly against a massive `composer update`, especially in a production environment, as all the other librairies will also potentially be updated without prior checking.


tabacitu:
Totally! Thank you!

PR #26, opened 2018-12-10
Fix wrong name of 'edit' button on the 'line' stack

aand18:
The button name is 'update', not 'edit'.

```
$this->crud->removeButton('edit', 'line');      // this doesn't remove the button
$this->crud->removeButton('update', 'line');    // this does
```


tabacitu:
Totally! Thank you for helping us make the docs better @aand18 !

PR #29, opened 2018-12-31
Email Column Documentation

AbbyJanke:
Email column type was not documented so i went to create my own and when doing PR to CRUD was already there xD


tabacitu:
Hmm... I also didn't know that was missing :-) Thank you @AbbyJanke ! Merged.

PR #30, opened 2019-01-05
Update crud-buttons.md

laxsmi:
I would suggest to update this since the button to create a record is now called "create"


tabacitu:
Totally! Thank you @laxsmi !

PR #32, opened 2019-01-15
Add few description to customize views.

faizfizy:
It helps me when I saw someone mention it in stackoverflow. So I think it's helpful so I add it here.


tabacitu:
Great, you're totally right - that wasn't clear, thank you @faizfizy !

PR #36, opened 2019-01-30
changed route call on function moderate instead of ban

aleszatloukal:


tabacitu:
Totally! Thank you @aleszatloukal , merged!

PR #38, opened 2019-02-19
Fix info for bootstrap datepicker, todayBtn don't accept true

pxpm:


tabacitu:
Great! Thank you @pxpm !

PR #39, opened 2019-04-03
Update the Laravel 5.7 install link

jsvini:


tabacitu:
You're right, thank you @jsvini !

PR #41, opened 2019-05-16
add doc about default value for select2

dendevs:
Actually documentation don't mention the value option for select2


tabacitu:
You're right. Thank you @dendevs , merged. Actually, it's recommended you use ```default``` instead of ```value```: 
- ```$field['value']``` gets populated from the db
- ```$field['default']``` is a fallback in case there's nothing in the db (or no entry whatsoever)

But I guess ```value``` will work too :-)

Cheers! Thanks for contributing!

PR #57, opened 2019-09-28
Feature/replace with extra routes

Patabugen:
During my upgrade I noticed that with() has been removed, so I've added as step.

Due to my merging mistake, it's also got a minor typo correction in 3.4. I can separate them if you like.


tabacitu:
Indeed - I totally forgot about eliminating ```->with()``` in routes. I personally never used it, preferred a normal route just under that one. Thank you!

PR #60, opened 2019-09-29
Match Required Laravel With Actual Crud Composer

Mu9tafa-Online:


tabacitu:
Hmm... actually I'm not sure we _do_ support 5.8. I though I dropped 5.8 support before launching. I'll check and get back to you. But thank you!

Mu9tafa-Online:
@tabacitu .. I Think Supporting Just **Laravel v6.0+** is Great.

It's already an LTS version + **PHP** is evolving everyday.

tabacitu:
Just checked - you're right, we did NOT drop support for L5.8, so your change makes total sense. Merged, thank you @Mu9tafa-Online !

PR #61, opened 2019-10-01
CRUD routes no longer have the crud. prefix

Patabugen:
The routes generated have lost their `crud.` prefix.

I believe this happened in v4 - I can't see it having changed in other recent versions.


Patabugen:
After publishing this I noticed the headings error in #62 that change is merged into this change already (so I'd suggest pulling that PR first)

tabacitu:
Totally. Thank you @Patabugen !

I can't, for the life of me, remember why I decided to remove the "crud" prefix though... And now I think it was a dumb idea... There was a much better separation with the crud prefix. Discussion here - https://github.com/Laravel-Backpack/CRUD/issues/2122

PR #63, opened 2019-10-02
include extra parameter for upload fields

zachweix:
Include that you can make temporary URL's for upload and upload multiple fields for files already on S3


tabacitu:
Totally. Thank you @zachweix !

PR #64, opened 2019-10-04
Fixes syntax error in the crud tutorial page

MartijnvAdrichem:
Removes the Admin namespace, because right now the url would be Admin\Admin\TagCrudController which is not what you want!


tabacitu:
Totally! Thank you @MartijnvAdrichem !

PR #69, opened 2019-11-11
Clarify about upload field validation rules.

pxpm:
The way backpack is designed, when an upload field is present and required if you submit a form without uploading the file an empty field is going to validate and fail ofcourse in the required rule. 

If you submit the form and everything is correct the process will be smooth, until you edit that entity.

When editing an entry if the upload field value is present (user already sent one file with success) we will just re-create the fancy file preview and not the upload field. If user submits the form without changing anything and we are not using **sometimes|required**, the validation will fail, as we are not sending any upload field in request. Here comes the power of [sometimes laravel validation rule](https://laravel.com/docs/5.8/validation#conditionally-adding-rules) that will not trigger the required validation error because the upload field is not present in request.

If user clicks to remove the file in the preview we re-create the empty input, so if the user sets a new file it will be uploaded and pass the validation, if don't send new file the validation will fail.

I think this is just mandatory if your upload is required, so it should be mentioned and clear in the docs.


tabacitu:
Totally! Thank you @pxpm !

PR #74, opened 2019-11-21
missing "function" text added

siberfx:


tabacitu:
Totally! Thank you @siberfx !

PR #79, opened 2019-12-15
unused [/block] tags removed

siberfx:
unused [/block] tags removed


tabacitu:
Right! Thank you @siberfx !

PR #80, opened 2019-12-16
Update bulk delete namespace import

fraserr:


tabacitu:
Right! Thank you @fraserr !

PR #85, opened 2020-01-15
Added example of Widget View name prefix for using ViewComposers

Patabugen:
It took me a while to figure it out


tabacitu:
Yup, I agree, the more examples the better. Sorry for the frustration @Patabugen ! Thanks!

PR #86, opened 2020-01-16
Update getting-started-basics.md added name field to migration schema.

SiCoUK:
Update artisan migration schema as it was missing the name field referred to later in the guide.

### DIFF
--- a/4.0/getting-started-basics.md
+++ b/4.0/getting-started-basics.md
@@ -56,7 +56,7 @@ This is where it gets interesting. As soon as you [install Backpack](/docs/{{ver
 
 ```zsh
 # STEP 1. create migration
-php artisan make:migration:schema create_tags_table --model=0 --schema="name:string:unique"
+php artisan make:migration:schema create_tags_table --model=0 --schema="slug:string:unique, name:string"
 php artisan migrate
 
 # STEP 2. create crud


tabacitu:
Hi @SiCoUK ,

I agree that would be more useful in a real-world setting. But for the absolute "basics" I think just one simple column is easier to understand. 

When you introduce the concept of "slug" other questions pop up:
- will you allow the user to edit the slug afterwards?
- is the slug auto-generated or does the user have to type it?

Stuff like that, that I would rather the reader not worry about yet.

**Wouldn't you agree? What made you want to submit this PR, did it feel incomplete?**

tabacitu:
PS. Thanks a lot for the PR. Really appreciate it.

SiCoUK:
@tabacitu I agree to a certain extent yes, the tutorial is not to build a real world tagging application and that is not what I was expecting.

However the reason I brought it up is because step 2 of the tutorial starts with the line "Let’s bring back the example in our first lesson. The Tags CRUD:" and then everything that follows includes a slug. Therefore while following along with the tutorial my demo application failed!

But don't get me wrong I am not complaining just trying to keep it consistent for someone coming in fresh.

tabacitu:
Oh I understand now. Thanks @SiCoUK ! I see we've fixed this in the meantime - https://backpackforlaravel.com/docs/4.0/getting-started-basics#cruds both name and slug are there, and unique.

So I'm going to close the PR. Thanks for reporting this though! It's difficult to spot these things ourselves, since we rarely run them. Cheers!

PR #97, opened 2020-02-08
add extra param to the image column

Yarob50:
I have been really struggling to find out a way to make an absolute path for images columns since I host my images in an s3 bucket, and it took me too long looking at the docs and in the internet trying to find a way to do that but I didn't. Finally, I decided to go to columns/image.blade.php and make my changes there, but it was surprising to me when I figured that there is already an option that you can add to the image field which is ```disk``` and it does exactly what I wanted but this wasn't mentioned in the docs, that's why I decided to add it there in case anyone was looking to do the same thing.

### DIFF
--- a/4.0/crud-columns.md
+++ b/4.0/crud-columns.md
@@ -232,6 +232,8 @@ Show a thumbnail image.
    'label' => "Profile image", // Table column heading
    'type' => 'image',
     // 'prefix' => 'folder/subfolder/',
+    // image from a different disk (like s3 bucket)
+    // 'disk' => 'disk-name', 
     // optional width/height if 25px is not ok with you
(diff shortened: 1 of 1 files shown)


tabacitu:
Totally! Thank you for contributing @Yarob50 - it's easy for the people who a feature to miss things in the documentation, because often times we _don't look at the docs_ (already know it by heart). 

I know this looks like a small PR from the outside, but these docs PRs help a lot - thank you.

Yarob50:
@tabacitu many thanks for your support, BTW this my first contribution and I am very happy that my PR has been merged and it's now reflected in the official docs. Backpack is amazing and I am sure that this not going to be my last contribution on it.

PR #106, opened 2020-03-11
Update image upload example

pxpm:
Show developers how they could remove any previous image from disk when uploading a new one.


tabacitu:
Hmmm are you sure this would work? Would the attribute then be the _current_ image?

I believe in order to delete the previous image this should be done _before_ step 0, not as step 4.

Am I mistaken?

pxpm:
Indeed, not mistaken you are right!

PR #108, opened 2020-03-13
Fixed publish command

genesiscz:
Docs have a bad command stated. 
```
$ php artisan backpack:crud:publish buttons/create                                            
  There are no commands defined in the "backpack:crud" namespace.  
  Did you mean this?                                               
      backpack                                                     
                                                                  
$ php artisan backpack:publish buttons/create
Cannot find source view file at vendor/backpack/crud/src/resources/views/buttons/create.blade.php - make sure you've picked an existing view file
```


tabacitu:
Ouch! You're right @genesiscz , thanks a lot for sending this. I've searched and found this problem across the docs, and fixed it everywhere. This must have been so annoying.

Cheers!

PR #115, opened 2020-04-24
Adding $id parameter in "publish"

flartet:
If I'm right (I've following the tuto), you need the $id parameter in the publish method to keep it related to the route where $id is present.
For someone who is mastering Laravel I think it's clear, but maybe you could, somewhere, add a note about putting parameters in route <=> public function ACTION($parameter)
Also, cherry on the cake, maybe a generator with optional parameter like
```
php artisan backpack:crud-operation --params=id Comment
php artisan backpack:crud-operation --params=key1,key2 Comment
```
2nd version to match composed primary key maybe

Thanks for the good work again.

### DIFF
--- a/4.0/crud-operations.md
+++ b/4.0/crud-operations.md
@@ -273,7 +273,7 @@ To add an operation to an ```EntityCrudController```, you can:
 ```
 - create a method that performs the operation you want:
 ```php
-    public function publish()
+    public function publish($id)
     {
         // do something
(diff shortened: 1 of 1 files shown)


tabacitu:
Totally! Thanks @flartet  - merging to ```dev``` for now, and it'll be publish on Monday most likely.

A generator command with parameters sounds sensible!

PR #116, opened 2020-04-25
page_or_link name field

maig81:
The 'name' of page_or_link field is somewhat misleading as it does not update the page_id or link fields.


tabacitu:
You're right! Thank you @maig81 - we must have missed that when we make names accept arrays in 4.0 - sorry for the frustration this must have caused you 🙄 Thanks for the PR!

PR #122, opened 2020-05-03
Removed not supported helper starts_with()

urlportal:
Fix does not need a description


tabacitu:
Totally. Thanks @urlportal .

PR #143, opened 2020-05-21
Fixed elfinder resources path in upgrade guide

cplewnia:


tabacitu:
Thank you @cplewnia !

PR #145, opened 2020-05-24
Removed obsolete upload attribute from image field

cplewnia:
Apparently the 'image' field no longer uses the 'upload' attribute. At least the corresponding view 'image.blade.php' does not make use of it anymore.

### DIFF
--- a/4.1/crud-fields.md
+++ b/4.1/crud-fields.md
@@ -748,7 +748,6 @@ $this->crud->addField([
     'label' => "Profile Image",
     'name' => "image",
     'type' => 'image',
-    'upload' => true,
     'crop' => true, // set to true to allow cropping, false to disable
     'aspect_ratio' => 1, // ommit or set to 0 to allow any aspect ratio
     // 'disk' => 's3_bucket', // in case you need to show images from a different disk


tabacitu:
Indeed @cplewnia the `upload` attribute isn't used inside the blade view. But it _is_ used when adding the field. ```Fields::hasUploadFields()``` takes into consideration if that's present on any fields, in order to make the form multipart, so uploads work. So I still think it's useful to have it.

Were you able to upload an image _without_ that, and without any other fields that have `upload` true? If you have, I might be wrong on this one - it's been known to happen 😆 

Thanks for opening this. Cheers!

cplewnia:
I think the issue is a bit more complicated. It depends on how the upload is implemented. Using a plain HTML form with a `file` type input field, we apparently need to set the form's encoding to `enctype="multipart/form-data"` in order to treat the file data as binary content.

However, the backpack `image` field uses JavaScript to get and process the file contents before submitting the form. In this processing the image is loaded as a base64 encoded string. This is done by ` fileReader.readAsDataURL(file);` in `vendor/backpack/crud/src/resources/views/crud/fields/base64_image.blade.php` and `vendor/backpack/crud/src/resources/views/crud/fields/image.blade.php`:

```javascript
$uploadImage.change(function() {
    var fileReader = new FileReader(),
            files = this.files,
            file;

    if (!files.length) {
        return;
    }
    file = files[0];

    const maxImageSize = 41943040;
    if(maxImageSize > 0 && file.size > maxImageSize) {

        alert(`Please pick an image smaller than ${maxImageSize} bytes.`);
    } else if (/^image\/\w+$/.test(file.type)) {

        fileReader.readAsDataURL(file);
        fileReader.onload = function () {

            $uploadImage.val("");
            $previews.show();
            if(crop){
                $mainImage.cropper(options).cropper("reset", true).cropper("replace", this.result);
                // Override form submit to copy canvas to hidden input before submitting
                // update the hidden input after selecting a new item or cropping
                $mainImage.on('ready cropstart cropend', function() {
                    var imageURL = $mainImage.cropper('getCroppedCanvas').toDataURL(file.type);
                    $hiddenImage.val(imageURL);
                    return true;
                });
                $rotateLeft.click(function() {
                    $mainImage.cropper("rotate", 90);
                });
                $rotateRight.click(function() {
                    $mainImage.cropper("rotate", -90);
                });
                $zoomIn.click(function() {
                    $mainImage.cropper("zoom", 0.1);
                });
                $zoomOut.click(function() {
                    $mainImage.cropper("zoom", -0.1);
                });
                $reset.click(function() {
                    $mainImage.cropper("reset");
                });
                $rotateLeft.show();
                $rotateRight.show();
                $zoomIn.show();
                $zoomOut.show();
                $reset.show();
                $remove.show();

            } else {
                $mainImage.attr('src',this.result);
                $hiddenImage.val(this.result);
                $remove.show();
            }
        };
    } else {
        new Noty({
            type: "error",
            text: "<strong>Please choose an image file</strong><br>The file you've chosen does not look like an image."
        }).show();
    }
});
```

As a consequence we are dealing with a regular string without any special characters. Thereby we do not profit from the `multipart/form-data` encoding set in `resources/views/vendor/backpack/crud/edit.blade.php`:

```php
@if ($crud->hasUploadFields('update', $entry->getKey()))
enctype="multipart/form-data"
@endif
```

Thus as long as the images are always base64 encoded, I assume the form tag does not need `enctype="multipart/form-data"` and thus the backpack field does not need the `upload` attribute. However, I have read that base64 encoding has thedownside of increasing the file size size. Thus it would be great to have the option to actually choose how the image is uploaded and derive the encoding from that.  This would require a revision of the Javascript image change handler implementation.

tabacitu:
Oh right! I guess you're right @cplewnia , we don't actually need `upload` here. Merged, thank you!

PR #150, opened 2020-06-07
Add note to date and datetime CRUD column documentation re: ISO formats

tibbsa:
After scratching my head for a while trying to figure out why date formats in my CRUD list view were not working as expected, I dove into the code and discovered that the Blade template that formats the columns is calling ```Carbon::isoFormat()``` rather than ```Carbon::format()``` (which I had been using elsewhere).  This changes the format strings that must be used.  

This patch simply adds a  note about this and provides a link to the Carbon docs describing the available format specifiers for ```isoFormat()```.

### DIFF
--- a/3.4/crud-columns.md
+++ b/3.4/crud-columns.md
@@ -174,7 +174,9 @@ Show custom HTML based on a closure you specify in your EntityCrudController. Pl
 ### date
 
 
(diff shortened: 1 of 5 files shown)


tabacitu:
Totally = much clearer now. Thank you for fixing this!

PR #151, opened 2020-06-08
Custom CSS/JS paths

martijnb92:


tabacitu:
Related to https://github.com/Laravel-Backpack/CRUD/issues/2923

Not sure if we should merge this PR and be done with it (assuming nobody has used this in 4.1) or implement a fallback mechanism like described in https://github.com/Laravel-Backpack/CRUD/issues/2923#issuecomment-655993591

I could use some feedback on this 😄

PR #152, opened 2020-06-09
Fix reference to CrudTrait

olipayne:


tabacitu:
Yup, totally. Thanks @olipayne !

PR #153, opened 2020-06-11
Backpack's Sluggable saves as JSON

dandarie:
Make it clear that Backpack's Sluggable always saves the slug as JSON, ignoring the translatable property.


tabacitu:
Totally! Thank you @dandarie - much better now.

PR #156, opened 2020-07-01
Add optional attribute 'inline' in code example of radio field

themagnifico:


tabacitu:
Thank you @themagnifico but I think you were looking at the COLUMNS docs 😄 The FIELDS docs already have the `inline` attribute specified 😆 https://backpackforlaravel.com/docs/4.1/crud-fields#radio

I see the PR is directed towards columns too - I think that's what was going on 😉 

Cheers!

tabacitu:
PS. Thanks for taking the time to submit the fix though - we really appreciate it.

PR #159, opened 2020-07-17
Misleading hasMany() relationship.

ChuiSauCE:
hasMany() relationship in Laravel refers to 1-n and doesn't require pivot table. While updating and creating hasMany() relationship in your model, this should be handled manually as explained in 'https://github.com/Laravel-Backpack/CRUD/issues/3031#issuecomment-660034201'. Referring this method through the documentation is misleading and better to be removed in my opinion.

There is also a relationship mentioned belongsToOne() which doesn't exist in Laravel core, removed this as well.

### DIFF
--- a/4.1/crud-fields.md
+++ b/4.1/crud-fields.md
@@ -1017,7 +1017,7 @@ Input preview:
 <a name="relationship"></a>
 ### relationship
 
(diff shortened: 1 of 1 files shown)


tabacitu:
Totally agree - thanks a lot for improving this bit @ChuiSauCE !

PR #160, opened 2020-07-17
Backticks changed to apostrophes in code examples

tomsb:
Backticks changed to apostrophes in code examples


tabacitu:
Totally! Now the code examples aren't useless, thanks @tomsb . This must've been from back when we were using readme.io - it converted the backticks sometimes.

Cheers!

PR #161, opened 2020-07-18
Fix two typos

sniperwolf:


tabacitu:
Totally, so annoying, thank you @sniperwolf for spotting and fixing this.

PR #164, opened 2020-07-21
Update crud-buttons.md

DorianLsl:
I got the error "PNotify is not defined" following the current doc.


tabacitu:
Indeed, we're using Noty instead of PNotify in Backpack 4.x - thanks for the fix @DorianLsl !

PR #170, opened 2020-08-13
Change 'category' to 'city'

yabbou:
I believe category was for the non-multiple select ajax example. I am also assuming it should be in the singular.


tabacitu:
Right! Totally, thank you @yabbou !

PR #172, opened 2020-08-27
Update upgrade-guide.md

paperscissors:
Give better idea of what needs to be done instead of using `parent::`


tabacitu:
Right! Much better, thank you @paperscissors !

PR #180, opened 2020-09-12
Problem with namespace in fetch operation

aula2030:
While trying the hasMany relationship, I realized that the select was not populated:

fetch-operation

The logs shown an error 500: `"POST /admin/proyecto/fetch/tareas HTTP/1.1" 500 84`

Problem was App\Models was concatenated to namespace so it needs a heading backslash. :+1:  So other people don't have to spend some time figuring it out.

Also, when I read the documentation, maybe because English is not my first language, I misunderstood. After examples of not-ajax and ajax, where it says:

> Then, you need to create the route and method that allows select2 to search and fetch the results...

I misunderstood and though that only the ajax example needed the fetch operation. Maybe you could change it to:

> In both cases, you need to create the route and method that allows select2 to search and fetch the results...

Cheers!!
X.

### DIFF
--- a/4.1/crud-fields.md
+++ b/4.1/crud-fields.md
@@ -1112,7 +1112,7 @@ Then, you need to create the route and method that allows ```select2``` to searc
 
     public function fetchCategory()
     {
-        return $this->fetch(App\Models\Category::class);
+        return $this->fetch(\App\Models\Category::class);
     }
 ```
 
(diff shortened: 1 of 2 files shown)


tabacitu:
Ugh, I imagine how frustrating it must have been for that backslash @aula2030 . Thanks a lot, merged!

PR #181, opened 2020-09-16
Update crud-fields.md

ikkebra:
Inside the fetchCategory function in the relationship field type docs, the leading backslash is required in order to get ajax to work.
i.e.
replace
return $this->fetch(App\Models\Category::class);
with
return $this->fetch(\App\Models\Category::class);


ikkebra:
My bad, this is a dupe of #180

PR #184, opened 2020-09-29
WIP - add fields FAQ - f

tabacitu:
The first FAQ details all relationships by type, which field you should use.
It's a breakdown of all relationships in Laravel - which field you should use for each relationship type.

Now that I'm working on this, I can't believe we've had 4-6 years without a breakdown line this 😞 it makes so much sense.

Thanks @pxpm 

Work in progress. Nowhere near finished.

### DIFF
--- a/4.1/crud-fields.md
+++ b/4.1/crud-fields.md
@@ -2170,3 +2170,120 @@ Inside your custom field type, you can use these variables:
 If your field type uses JavaScript, we recommend you:
 - put a ```data-init-function="bpFieldInitMyCustomField"``` attribute on your input;
 - place your logic inside the scripts section mentioned above, inside ```function bpFieldInitMyCustomField(element) {}```; of course, you choose the name of the function but it has to match whatever you specified as data attribute on the input, and it has to be pretty unique; inside this method, you'll find that ```element``` is jQuery-wrapped object of the element where you specified ```data-init-function```; this should be enough for you to not have to use IDs, or any other tricks, to determine other elements inside the DOM - determine them in relation to the main element; if you want, you can choose to put the ```data-init-function``` attribute on a different element, like the wrapping div;
+
+
+<a name="frequently-asked-questions-about-fields"></a>
+## FAQs
+
+<a name="what-field-should-i-use-for-a-relationship"></a>
+### What field should I use for a relationship?
+
+With so many field types, it can be a little overwhelming for a first-timer to quickly grasp what field type to use for your Eloquent relationship. Here's a list of what most people use, most of the times:
+
+#### hasOne (1-1 relationship)
+
+- example: 
+    - `User -> Phone`
+    - a User has one Phone; a Phone can only belong to one User
+    - the foreign key is stored on the Phone (`user_id` on `phones` table)
+- how to use:
+    - [the `hasOne` relationship should be properly defined](https://laravel.com/docs/eloquent-relationships#one-to-one) in the User model;
+    - you can easily add fields for each individual attribute on the related entry; you just need to specify in the field name that the value should not be stored on the _main model_, but on _a related model_; you can do that using dot notation (`relationship_name.column_name`); note that the prefix (before the dot) is the Model name, not the Table name;
+    - all fields types should work fine - depending on your needs you could choose to add a `text` field, `number` field, `select` field, etc;
+
+```php
+// inside UserCrudController::setupCreateOperation()
+CRUD::field('phone.number')->type('number');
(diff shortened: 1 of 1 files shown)


4.1/crud-fields.md
```diff
@@ -2170,3 +2170,127 @@ Inside your custom field type, you can use these variables:
...
+
+// TODO
+
+#### morphMany (1-n polymorphic relationship)
+
+// TODO
+
+#### morphToMany (n-n polymorphic relationship)
+
+// TODO
```
tabacitu, line 2296: We should make sure this new section includes ALL Laravel relationships, so that we leave no question unanswered. Whatever Eloquent relationship the dev wants to use, they'd find a corresponding field here.

@pxpm could you please take a look at these extra relationships, and let me know the fields that best make sense for each one? What you did in that email was WONDERFUL.

4.1/crud-fields.md
```diff
@@ -2170,3 +2170,127 @@ Inside your custom field type, you can use these variables:
...
+// inside RoleCrudController::setupCreateOperation()
+CRUD::field('users')->type('relationship');
+CRUD::field('users')->type('select_multiple');
+CRUD::field('users')->type('select2_multiple');
+```
+
+- notes:
+    - if you choose the `relationship` field, you can also use [the InlineCreate operation](/docs/{{version}}/crud-operation-inline-create), which will add a [+ Add Item] button next to the dropdown, to let the admin create a Role in a modal, without leaving the current Create User form; 
+
+// TODO: what if you want to save extra information on the pivot table?
```
tabacitu, line 2274: We currently don't have a good solution for storing extra info on the pivot table, do we?

AFAIK it's _possible_, using the `repeatable` field type and overwriting the `store()` and `update()` methods a little bit, but we don't have instructions on how to do that either...

4.1/crud-fields.md
tabacitu: This will probably address https://github.com/Laravel-Backpack/CRUD/issues/1926

tabacitu:
@pxpm I propose we fill in the docs with what we support right now in Backpack 4.1, and be honest and tell people there is no easy way to do it (or tell the complicated way) where needed. That way, we have all relationships documented in 4.1.

Then when we duplicate the docs for 4.2, we just replace the "not implemented" sections with the right instructions.

What do you think?

4.1/crud-fields.md
```diff
@@ -2170,3 +2170,127 @@ Inside your custom field type, you can use these variables:
...
+
+#### hasOne (1-1 relationship)
+
+- example: 
+    - `User -> Phone`
+    - a User has one Phone; a Phone can only belong to one User
+    - the foreign key is stored on the Phone (`user_id` on `phones` table)
+- how to use:
+    - [the `hasOne` relationship should be properly defined](https://laravel.com/docs/eloquent-relationships#one-to-one) in the User model;
+    - you can easily add fields for each individual attribute on the related entry; you just need to specify in the field name that the value should not be stored on the _main model_, but on _a related model_; you can do that using dot notation (`relationship_name.column_name`); note that the prefix (before the dot) is the Model name, not the Table name;
```
pxpm, line 2191:
```suggestion
    - you can easily add fields for each individual attribute on the related entry; you just need to specify in the field name that the value should not be stored on the _main model_, but on _a related model_; you can do that using dot notation (`relationship_name.column_name`); note that the prefix (before the dot) is the **Relation** name, not the table name;
```

pxpm:
@tabacitu please give this a look. 

I tested all examples I created in gists so I am 100% sure they are working with the current 4.1 version. If anything goes wrong is probably a typo when copying to github interface and adapting the names 🙃

Man .. I will tell you what .. I've been working with the branch I created with all this relationship quirck fixes, and setting this up "vanilla" has taken me so long! things are about to get WAY WAY easier in 4.2. 

We still have some stuff to discuss if we should do in 4.1 like the `morph => true` we already allow somehow in fields, but it can only work for `morphToMany` not the other morphs. Or the `pivotFields` for the `belongsToMany` that I found no way of make it work? 

You can't imagine how happy I am that I submitted this PR. Sure, some stuff to fix, but the heavy lifting is done for good! For 4.2 I hope it's just a matter of erasing stuff we will not need anymore!

tabacitu:
This. Is. BONKERS! Excellent job @pxpm , I have zero comments 🎉🎉🎉 I haven't tried your gists, just gave them a quick look, but if they work this is... long overdue 😅 I love this! ❤️

slow_clap_cap

PR #186, opened 2020-10-06
Update installation text for Laravel 8 users

mahasadhu:
Since https://github.com/Laravel-Backpack/CRUD/pull/3241 was merged, I think the docs need to be updated too. 

Update installation text to tell backpack users to update `User` model in config if they are using Laravel 8 and up.


tabacitu:
Totally! Thanks a lot @mahasadhu .

tabacitu:
Also, I just submitted https://github.com/Laravel-Backpack/CRUD/pull/3269 which should solve this so that "it just works". If you could take a look, that would help us out. If you can't, we don't mind 😄 of course.

Cheers!

PR #187, opened 2020-10-11
Update custom store operation method example

danilopolani:
The code has been taken from the [PackageManager code](https://github.com/Laravel-Backpack/PermissionManager/blob/master/src/app/Http/Controllers/UserCrudController.php#L103-L124) (as per linked in the docs).

### DIFF
--- a/4.1/crud-fields.md
+++ b/4.1/crud-fields.md
@@ -986,14 +986,20 @@ public function setPasswordAttribute($value) {
     
     public function store()
     {
-        $this->crud->request = $this->crud->validateRequest();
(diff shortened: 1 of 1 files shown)


tabacitu:
Totally! 

Thank you @danilopolani - I know people have complained about this bad example before but I must have missed it here. Thanks a lot for fixing it - and sorry for the frustration, I imagine it was annoying to see the example didn't work 🙄 😄 

Cheers!

danilopolani:
No problem! In a big (great) project like this an help must come from users too :)

PR #193, opened 2020-11-11
Add .show() to Noty to render alert in crud-buttons.md

s-nunez:


tabacitu:
Thanks @s-nunez . Must have been leftover from 4.0, the previous version of Noty didn't need that.

PR #201, opened 2020-11-23
preventing fetch from infering searchable attributes

pxpm:


4.1/crud-operation-fetch.md
```diff
@@ -64,6 +64,29 @@ class ProductCrudController extends CrudController
 
 Based on the fact that the ```fetchTag()``` method exist, the Fetch operation will create a ```/product/fetch/tag``` POST route, which points to ```fetchTag()```. Inside ```fetchTag()``` we call ```fetch()```, that responds with entries in the format ```select2``` needs.
 
+**Preventing FetchOperation of guessing the searchable attributes**
+
+If not specified `searchable_attributes` will be automatically infered from model dabatase columns. To do your own search and replace default FetchOperation behaviour you can setup an empty `searchable_attributes`.
```
tabacitu, line 69: @pxpm I'm afraid I don't understand 😅 Search&replace? What replace?

Could you please rephrase in more words so I understand when this is needed? Happy to help make this easier to understand 😀

tabacitu (changes requested):

4.1/crud-operation-fetch.md
pxpm, line 69:
```suggestion
If not specified `searchable_attributes` will be automatically infered from model dabatase columns. To prevent this behaviour you can setup an empty `searchable_attributes` array.
```

pxpm: @tabacitu this is usefull for when you want to write your own query instead of the default FetchOperation query. 

Without this knowledge don't setting up `searchable_attributes` we will infer from the "possible names" if not found, the first column in database. 

Doing this we apply atleast one constrain to the query when using FetchOperation. Setting up an empty `searchable_attributes` will make Fetch not guess from name list or database column.

PR #211, opened 2020-12-08
add how to fetch a select2_ajax filter

pxpm:


4.1/crud-operation-fetch.md
```diff
@@ -87,6 +87,27 @@ public function fetchUser() {
     }
 ```
 
+## Using FetchOperation with `select2_ajax` filter
```
pxpm, line 90:
```suggestion
<a name="fetch-ajax-filter"></a>
## Using FetchOperation with `select2_ajax` filter
```

4.1/crud-operation-fetch.md
```diff
@@ -87,6 +87,28 @@ public function fetchUser() {
...
+
+```
+$this->crud->addFilter([
+  'name'        => 'category_id',
+  'type'        => 'select2_ajax',
+  'label'       => 'Category',
+  'placeholder' => 'Pick a category',
+  'method' => 'POST', //mandatory change
+  'select_attribute' => 'name' //the attribute that will be shown to the user by default 'name'
+  'select_key' => 'id' //by default is ID, change if your model use some other key
```
promatik, line 103:
```suggestion
  'method' => 'POST', // mandatory change
  // 'select_attribute' => 'name' // the attribute that will be shown to the user by default 'name'
  // 'select_key' => 'id' // by default is ID, change it if your model uses some other key
```

Should we let the default attributes commented? To encourage users to not use it if they are copy/pasting this code?

promatik (changes requested):

4.1/crud-operation-fetch.md
pxpm: 💯 agree.

promatik (approved):

4.1/crud-operation-fetch.md
tabacitu: Hmm... I wouldn't put these here. I think it'd be better if we put this inside the filter directly. As an either/or. For example: _as the source of your filter, you can either (A) use the FetchOperation or (B) code your own route, and controller method that responds with results_. Then detailing steps for the two sources of input.

The reason I'm thinking this is because... I don't think people will come to the FetchOperation and think "_I want to use this in a filter_". No, what's more likely is that they'll go _to the filters docs_ to add one, and when they get to `select2_ajax` they expect it to work with their existing FetchOperation. At least that's what I think.

So I think moving these instructions to filters will:
- make them available where the user is looking for them
- keep the FetchOperation docs isolated from where it's used (fields and filters); though we _could_ link to instructions to the instructions in both fields and filters, just in case people come here first;

tabacitu (changes requested):

4.1/crud-operation-fetch.md
```diff
@@ -87,6 +87,28 @@ public function fetchUser() {
...
+  'method' => 'POST', // mandatory change
+  // 'select_attribute' => 'name' // the attribute that will be shown to the user by default 'name'
+  // 'select_key' => 'id' // by default is ID, change it if your model uses some other key
+],
+url('admin/fetch/monster'), // the fetch route 
+function($value) { // if the filter is active
+    // $this->crud->addClause('where', 'category_id', $value);
+});
+
+```
```
tabacitu, line 110: Looking at this syntax, and how we'd use this filter + fetch, makes me wonder. Wouldn't it be more intuitive to have a new filter `fetch`, that does this without having "_mandatory changes_"? That might:
- have less mandatory parameters (because it can have different defaults)
- simplify the instructions: 
    - want to use your own code as a data source? use the `select2_ajax` filter;
    - want to use the fetch operation as a data source? use the `fetch` filter;

What do you think @pxpm ? The way I see it the `fetch` filter could be either:
- (A) copy-paste of the `select2_ajax` in our current PR (fixed) 
- (B) it could be a super-simple file where it sets different defaults, then includes the same `select2_ajax` blade file

I'd say (B) is cleaner.

The end result, I think, would be a cleaner syntax for adding a filter with the Fetch operation as its data source, right? Something like:
```php
$this->crud->addFilter([
   'name'        => 'category_id',
   'type'        => 'fetch',
   'label'       => 'Category',
   // 'placeholder' => 'Pick a category',
   // 'method' => 'POST', // mandatory change
   // 'select_attribute' => 'name' // the attribute that will be shown to the user by default 'name'
   // 'select_key' => 'id' // by default is ID, change it if your model uses some other key
 ],
 url('admin/fetch/monster'), // the fetch route 
 function($value) { // if the filter is active
     // $this->crud->addClause('where', 'category_id', $value);
 });

// which fluently would look like this
CRUD::filter('category_id')->label('Category')->type('fetch')
    ->options(url('admin/fetch/monster'))
    ->whenActive(function($value) {
        CRUD::addClause('where', 'category_id', $value);
    });
```

What do you think @pxpm ? Worth separating... or overkill/oversimplification?

promatik: I'll give my thoughts on this, a `fetch` filter would be nice in that way, it would be easier to set up, we could for instance shorten the route, for `url('admin/monster/fetch/icon')` we know already `url('admin/monster/fetch/')` the missing part is just `icon`, we could let developers provide just that for instance.

But, I think it's a overkill, we have already many filters; `select2_multiple`, `select2_ajax`, `select2`, `dropdown` and we have `select2_multiple_ajax` to do.

**I think** the way is to do what was done with fields, a `relationship` filter (or just select2), where you add `multiple` attribute or not, and most of the attributes like source are automated.
(_This is controversy 😅 do you guys agree with this? @tabacitu, @pxpm?_)

So, creating another filter is just a small step backwards in my opinion.

tabacitu: I agree. Let's add this feature to `select2_ajax` because it's the quickest thing to do. Then think about adding a new `relationship` filter in c4.2 or v12, when we maybe get to standardize stuff a little bit.

4.1/crud-operation-fetch.md
```diff
@@ -87,6 +87,28 @@ public function fetchUser() {
...
+$this->crud->addFilter([
+  'name'        => 'category_id',
+  'type'        => 'select2_ajax',
+  'label'       => 'Category',
+  'placeholder' => 'Pick a category',
+  'method' => 'POST', // mandatory change
+  // 'select_attribute' => 'name' // the attribute that will be shown to the user by default 'name'
+  // 'select_key' => 'id' // by default is ID, change it if your model uses some other key
+],
+url('admin/fetch/monster'), // the fetch route
```
tabacitu, line 105:
```suggestion
backpack_url('product/fetch/category'), // the fetch route on the ProductCrudController 
```

4.1/crud-operation-fetch.md
```diff
@@ -87,6 +87,28 @@ public function fetchUser() {
     }
 ```
 
+<a name="fetch-ajax-filter"></a>
+## Using FetchOperation with `select2_ajax` filter
+
+FetchOperation can also be used with `select2_ajax` filter. To make it work we need to setup aditional attributes in the filter. By default the `method` used by FetchOperation is `POST` while in filter is `GET`. We also need to tell backpack what attribute we want to show to the user.
```
tabacitu, line 93:
```suggestion
The FetchOperation can also be used as the source URL for the `select2_ajax` filter. To do that, we need to:
- change the AJAX method from `GET` (the default for this filter) to `POST` (the default for the Fetch operation);
- tell the filter what attribute we want to show to the user; 
```

PR #221, opened 2021-01-04
Update crud-operation-reorder.md

EricPaulson:
I noticed in my code that I need to return the trait to get this to work. Hopefully this will save someone some time later!


tabacitu:
I'm sure it will 😀 Thank you!

PR #223, opened 2021-01-08
Carbon instead of Date in the example casting code

sergiks:


tabacitu:
Totally! Thank you @sergiks !

PR #233, opened 2021-02-04
Update crud-operations.md

Mohammed-Elmasry:
Add <OperationName> in between HTML brackets to enhance readability and comprehension.

While reading the documentation, I didn't understand what was meant until I read too carefully and it clicked. As a programmer, I believe we should write self-explanatory code. We need to write code that reads like well-written prose. It has to explicitly tell the intentions of the developer who wrote it.

That's why I hope you update the Documentation.

### DIFF
--- a/4.0/crud-operations.md
+++ b/4.0/crud-operations.md
@@ -20,8 +20,8 @@ class ProductCrudController extends CrudController
 ```
 
 **Operations are traits that add functionality to that controller**. Most operations will have:
-- routes inside a ```setupOperationNameRoutes()```; this gets called in your ```routes/backpack/custom.php``` by the ```Route::crud('product', 'ProductCrudController``` macro, which determines which routes to register for that CrudController;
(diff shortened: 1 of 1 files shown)


promatik:
Thank you @Mohammed-Elmasry, having the operation name between `<` `>` is not amazingly beautiful 😅 but indeed it's more intuitive for developers 👌

---

image

Mohammed-Elmasry:
I do realize that @promatik. It is not the most elegant thing to do, but I cannot explain how much time I spent looking around in my CrudControllers for that function name :smile:

tabacitu:
Totally agree! Thanks for the suggestion @Mohammed-Elmasry ! 

Let us know if you find more things that annoy your like this in our docs. For us they're more difficult to detect since we've gotten used to them. We're open to improvement - sometimes we actually _can't_ see it the same way you do, until you tell us. So your perspective is very welcome!

Cheers!

Mohammed-Elmasry:
Oh thank you. @tabacitu I would add more actually. Hopefully I will be of good help to you.

PR #234, opened 2021-02-04
Update crud-columns.md

pxpm:


4.1/crud-columns.md
```diff
@@ -13,6 +13,8 @@ It's used inside default operations to:
 
 A column consists of only one file - a blade file with the same name as the column type (ex: ```text.blade.php```). Backpack provides you with [default column types](#default-column-types) for the common use cases, but you can easily [change how a default field type works](#overwriting-default-column-types), or [create an entirely new field type](#creating-a-custom-column-type).
 
+> IMPORTANT NOTE: Starting with Backpack 4.1, if the _column name_ is the exact same as a relation method in the model, Backpack will assume you're adding a column for that relationship and infer relation attributes from it. To disable this behaviour, you can use `'entity' => false` in your column definition.
```
tabacitu, line 16:
```suggestion
> IMPORTANT NOTE: Starting with Backpack 4.1, if the _column name_ is the exact same as _a relation method_ in the model, Backpack will assume you're adding a column for that relationship and infer relation attributes from it. To disable this behaviour, you can use `'entity' => false` in your column definition. 
```

tabacitu:
Pedro I just merged this but now I'm having second thoughts 😀

This page is meant to be read to understand the basics about columns - I now think it's premature to tell people about something that is 4.1-specific and entity-specific, at that point. They've just learnt what columns are.

I agree with the note, but... what would be a better place for it?

PR #243, opened 2021-03-23
fix charts widget example typo

MikeyBeLike:
Small code typos in the example docs


promatik:
Hi @MikeyBeLike!
Thank you for the fix, indeed it's incorrect, probably due to the strange formatting 😅

I'll move this to _Ready to Merge_, thank you 🙌

PR #246, opened 2021-04-02
Set image attribute when creating new class object

MerdanDev:
I had problem with SQL error, that image attribute doesn't have default value when seeding my my table. For example:
$obj = new \App\Models\Product();
$obj->name = 'name';
$obj->image = 'image path';
$obj->save();

### DIFF
--- a/4.1/crud-fields.md
+++ b/4.1/crud-fields.md
@@ -908,6 +908,10 @@ Class Product extends Model
             $public_destination_path = Str::replaceFirst('public/', '', $destination_path);
             $this->attributes[$attribute_name] = $public_destination_path.'/'.$filename;
         }
+	else
+        {
+            return $this->attributes['image'] = $value;
+        }
     }
     
 // ..


tabacitu:
Haha very nifty @MerdanDev , I like it. 
Indeed, I see no reason why this shouldn't also support adding images by path. 

@pxpm / @promatik do you guys foresee any problems with this? I don't.

4.1/crud-fields.md
```diff
@@ -908,6 +908,10 @@ Class Product extends Model
             $public_destination_path = Str::replaceFirst('public/', '', $destination_path);
             $this->attributes[$attribute_name] = $public_destination_path.'/'.$filename;
         }
+	else
+        {
+            return $this->attributes['image'] = $value;
```
promatik, line 913:
```suggestion
            return $this->attributes[$attribute_name] = $value;
```

4.1/crud-fields.md
```diff
@@ -908,6 +908,10 @@ Class Product extends Model
...
+        }
```
promatik, line 914:
```suggestion
        else
        {
            return $this->attributes[$attribute_name] = $value;
        }
```

promatik (changes requested):

promatik:
I don't think this will bring any trouble for us, at least I've been using this on my `trait SaveMedia` for a long time 😅

image

4.1/crud-fields.md
promatik: Hi @MerdanDev, thank you for this suggestion, I approve it 👌
I have just this small changes to apply, and I see you've deleted the repo from your side, so I cannot apply the changes 😅, can you please try to somehow recover it?

PR #255, opened 2021-07-22
Update fakes documentation for translatable attributes

ablunier:
Documentation for fakes was confusing because after some testing I came up finding out that for translatable attributes to work as fakes they can't be in `$casts`.


tabacitu:
Huh - wasn't aware of that. Haven't used translatable fake attributes in a while. Thank you @ablunier !

PR #262, opened 2021-09-20
Docs: Exempt attributes when cloning

tomhelmer:


4.1/crud-operation-clone.md
```diff
@@ -114,3 +114,14 @@ You can also overwrite the bulk clone button by creating a file with the same na
 ```zsh
 php artisan backpack:publish crud/buttons/bulk_clone
 ```
+
+<a name="exempt-attribute-when-cloning"></a>
+## Exempt attributes when cloning
+If you have attributes that should not be cloned eg. a SKU with an unique constraint you can overwrite the replicate method on your model:
```
tabacitu, line 120:
```suggestion
If you have attributes that should not be cloned (eg. a SKU with an unique constraint), you can overwrite the replicate method on your model:
```

tabacitu:
Thank you @tomhelmer . I'm sure someone will find this very useful!

PR #264, opened 2021-10-02
Update crud-fields.md

vandolphreyes:
Adding the missing multiple field, without it it won't work on the first try.


tabacitu:
Thanks for taking the time to submit this @vandolphreyes .

@pxpm , could you please take a look at this field and `select_multiple`? This might actually be a bug... it seems redundant to me to have to specify `multiple => true` for a field that is called `select2_multiple`... What do you think?

Issue #266, opened 2021-10-05
Add note that output is not escaped in preview for EasyMDE, SimpleMDE and Summernote


tabacitu:
In these three editors, when you click the preview button, the output is not escaped at all. So if there's some JS there (eg: `afafa"><img src=x onerror=confirm(1)>`) then that JS will be run.

Not a problem for one-admin systems. Since you can do that from the Chrome Developer Tools anyway.
Yes a problem for multi-admin systems. One admin can potentially run JS on anothers' machine. Or a malicious user even, if the input is in the db from a separate form (outside Backpack).

I say let's add notes to these fields, saying that the preview is unescaped, so they should _maybe_ sanitize the input before they store it in the database, which is easiest done using a Laravel mutator for that attribute.

PR #267, opened 2021-10-06
Add a note in WYSIWYG editors about content escaping

pxpm:
refs: #266

### DIFF
--- a/4.1/crud-fields.md
+++ b/4.1/crud-fields.md
@@ -751,7 +751,7 @@ Show an [EasyMDE - Markdown Editor](https://easy-markdown-editor.tk/) to the use
 ],
 ```
 
-> NOTE: The contents displayed in the editor are NOT escaped by default. If you do NOT trust the admins who have access to this field (or end-users can also store information to this db column), please make sure this attribute is always escaped, before it's shown. You can do that by running the value through `strip_tags()` in a [Laravel accessor](https://laravel.com/docs/8.x/eloquent-mutators#accessors-and-mutators) on the model. Here's [an example](https://github.com/Laravel-Backpack/demo/commit/509c0bf0d8b9ee6a52c50f0d2caed65f1f986385).
+> NOTE: The contents displayed in this editor are NOT stripped, sanitized or escaped by default. Whenever you store Markdown or HTML inside your database, it's HIGHLY recommended that you sanitize the input or output. Laravel makes it super-easy to do that on the model using [accessors](https://laravel.com/docs/8.x/eloquent-mutators#accessors-and-mutators). If you do NOT trust the admins who have access to this field (or end-users can also store information to this db column), please make sure this attribute is always escaped, before it's shown. You can do that by running the value through `strip_tags()` in an accessor on the model (here's [an example](https://github.com/Laravel-Backpack/demo/commit/509c0bf0d8b9ee6a52c50f0d2caed65f1f986385)) or better yet, using an [HTML Purifier package](https://github.com/mewebstudio/Purifier) (here's [an example](https://github.com/Laravel-Backpack/demo/commit/7342cffb418bb568b9e4ee279859685ddc0456c1)). 
 
 Input preview: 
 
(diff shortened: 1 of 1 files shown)


tabacitu:
@pxpm just had an idea about this. I think we _can_ do a reasonable job of preventing JS in markdown fields. Hear me out:
- for JS that is NOT stored in the DB yet, nothing we can do; nothing we _should_ do either since that's possible through Chrome DevTools too;
- but for JS that _is stored in the DB_ already, what I think we _can_ do is run a `strip_tags()` on the value, before we pass it to EasyMDE/SimpleMDE/SummerNote;

Yes, that would limit the markdown editors functionality - they can no longer type code inside the markdown and have it "just work". But I'd argue that shouldn't be enabled by default anyway.

So what I propose we do, in addition to this docs change, is make it _easy_ for devs to escape those fields, if they want to. How? A field flag, `escaped`, that:
- in v4.1 defaults to `false`; if set to `true` will `strip_tags()` the value before it's used by the editor;
- in v4.2 defaults to `true`;

That way we don't just add a notice, but we make it easy for the dev to adopt a "safer" behaviour. And we make it easy for ourselves to adopt that as a default in the next version.

What do you say @pxpm ? See any fault in my logic - am I missing something so should we just go ahead and do this?

4.1/crud-fields.md
```diff
@@ -751,6 +751,8 @@ Show an [EasyMDE - Markdown Editor](https://easy-markdown-editor.tk/) to the use
 ],
 ```
 
+> NOTE: Please take into consideration that the contents displayed in the editor are not escaped by default. You should take care about escaping the contents, or don't allow it to reach the database in the first place.
```
tabacitu, line 754:
```suggestion
> NOTE: The contents displayed in the editor are NOT escaped by default. If you do NOT trust the admins who have access to this field (or end-users can also store information to this db column), please make sure this attribute is always escaped, before it's shown. You can do that by running the value through `strip_tags()` in a [Laravel accessor](https://laravel.com/docs/8.x/eloquent-mutators#accessors-and-mutators) on the model. Here's [an example](https://github.com/Laravel-Backpack/demo/commit/509c0bf0d8b9ee6a52c50f0d2caed65f1f986385).

File: 4.1/crud-fields.md [hunk 2]
```diff
@@ -1771,6 +1773,7 @@ Show a [SimpleMDE markdown editor](https://simplemde.com/) to the user.
     // 'simplemdeAttributesRaw' => $some_json
 ],
 ```
+> NOTE: Please take into consideration that the contents displayed in the editor are not escaped by default. You should take care about escaping the contents, or don't allow it to reach the database in the first place.
```
tabacitu, line 1776:
```suggestion
```suggestion
> NOTE: The contents displayed in the editor are NOT escaped by default. If you do NOT trust the admins who have access to this field (or end-users can also store information to this db column), please make sure this attribute is always escaped, before it's shown. You can do that by running the value through `strip_tags()` in a [Laravel accessor](https://laravel.com/docs/8.x/eloquent-mutators#accessors-and-mutators) on the model. Here's [an example](https://github.com/Laravel-Backpack/demo/commit/509c0bf0d8b9ee6a52c50f0d2caed65f1f986385).
```

4.1/crud-fields.md
```diff
@@ -1806,6 +1810,8 @@ Show a [Summernote wysiwyg editor](http://summernote.org/) to the user.
 
 ```
 
+> NOTE: Please take into consideration that the contents displayed in the editor are not escaped by default. You should take care about escaping the contents, or don't allow it to reach the database in the first place.
```
tabacitu, line 1813:
```suggestion
> NOTE: Summernote does NOT sanitize the input. If you do not trust the users of this field, you should sanitize the input or output using something like HTML Purifier. Personally we like to use install [mewebstudio/Purifier](https://github.com/mewebstudio/Purifier) and add an [accessor or mutator](https://laravel.com/docs/8.x/eloquent-mutators#accessors-and-mutators) on the Model, so that wherever the model is created from (admin panel or app), the output will always be clean. [Example here](https://github.com/Laravel-Backpack/demo/commit/7342cffb418bb568b9e4ee279859685ddc0456c1).
```

tabacitu:
Ok so I've been on this for a while now, and I no longer agree with 2h-ago-me. I don't think we should provide an `escaped` attribute to these fields... I think we should instruct people to _properly sanitize their inputs and outputs_. That's because:
- even if the JS editor itself sanitizes input, a malicious actor can still use Chrome DevTools to alter the input value before it's saved; so whatever editors like CKEditor do to sanitize the input is welcome, but not sufficient;
- that should be done ANYWAY, if you store complex strings like HTML or Markdown in the database;
- Laravel already provides a pretty simple way to make sure your DB entries are stored/outputted securely - using accessors and mutators;
- there's this great package (https://github.com/mewebstudio/purifier) that does everything for you to sanitize HTML, you just have to use it (or alternatively for markdown, use `strip_tags()`);

So rather than telling people to use `escaped => true` which can only do a front-end job for the admin panel, **I think it's better to resolve this INSIDE THE MODEL, using _better documentation_**. That way we teach people to properly clean or sanitize their inputs and outputs, and we not only make their admin panel more secure, but their entire application.

I've done the best I can to improve the docs in this sense in this PR. Please let me know if you find a better way of instructing people to do this.

I've also opened [an issue inside that repo](https://github.com/mewebstudio/Purifier/issues/152) so that maybe we create a custom cast, to make it dead-simple for people to sanitize their inputs/outputs. The easier we make it, the more likely they actually will...

But for now, I think this matter is resolved, using docs. Let me know if you want to reopen/improve this Pedro.

PR #278, opened 2021-11-21
Deprecate closure column

promatik:
Related with https://github.com/Laravel-Backpack/CRUD/pull/3936.

### DIFF
--- a/4.2-dev/crud-columns.md
+++ b/4.2-dev/crud-columns.md
@@ -196,7 +196,7 @@ Show custom HTML based on a closure you specify in your EntityCrudController.
 ],
 ```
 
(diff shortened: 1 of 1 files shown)


4.2-dev/crud-columns.md
```diff
@@ -196,6 +196,8 @@ Show custom HTML based on a closure you specify in your EntityCrudController.
 ],
 ```
 
+> **DEPRECATED**: closure column will be removed in a future version of Backpack, the same thing can now be achieved using any column (including the `text` column) and the `value` attribute - just pass the same closure to the `value` attribute of any column type.
```
tabacitu, line 199:
```suggestion
> **DEPRECATED**: closure column will be removed in a future version of Backpack, since the same thing can now be achieved using any column (including the `text` column) and the `value` attribute - just pass the same closure to the `value` attribute of any column type.
```

Issue #296, opened 2021-12-30
Unclear fields api documentation for Create and Update


Polfo:
https://backpackforlaravel.com/docs/4.1/crud-fields#fields-api-1

shows

```php
// add a field to both Create and Update operation
$this->crud->addField($field_definition_array);

// add a field only to the Update operation
$this->crud->addField($field_definition_array);
```
That is two times the same code with different comment.
This should probably be

```php
// add a field to the Create and/or Update operation depending on the `setupOperation()` context
$this->crud->addField($field_definition_array);
```

tabacitu:
Solved by https://github.com/Laravel-Backpack/docs/pull/297

PR #297, opened 2021-12-30
Update crud-fields.md

Polfo:
Solves #296


tabacitu:
Definitely! Thank you @Polfo .

PR #298, opened 2021-12-30
Update crud-fields.md

Polfo:
Avoid use of 'should' while explaining optional fields.
Avoid use of 'normally' for something that is an option.
Use name, label, type sequence everywhere.


tabacitu:
Agreed - that's better, thank you!

PR #301, opened 2021-12-30
Update crud-fields.md

Polfo:
Fix html input field is not displayed.
Remove redundant documentation, consistent with other fields, already documented in https://backpackforlaravel.com/docs/4.1/crud-fields#optional-field-attributes-for-presentation-purposes-1

### DIFF
--- a/4.2-dev/crud-fields.md
+++ b/4.2-dev/crud-fields.md
@@ -802,7 +802,7 @@ Input preview:
 <a name="hidden"></a>
 ### hidden
 
-Include an <input type="hidden"> in the form.
+Include an `<input type="hidden">` in the form.
 
 ```php
 [   // Hidden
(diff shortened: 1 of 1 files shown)


tabacitu:
Good catch @Polfo , thank you. I agree to the "input type=hidden" change, didn't even know that was an issue, I merged your other PR doing that 👍

I agree that these docs are duplicated too, and that it's odd... but that was a deliberate choice. People kept asking "how do I do x" and "how do I do y", all of them things fields could do.... through documented attributes at the top of that page. As soon as we added these lines in the "text" field (which everybody uses)... no more questions 😀 Turns out a lot of people don't the whole "fields page", they just go straight into development. Which makes sense, the page is huge.

So I'd rather keep this duplication, since it helps remind both newbies and old-timers of the optional attributes.

Cheers!

PR #314, opened 2022-01-27
V5 docs

tabacitu:
Turns out we can't call this release v4.2, we _have to_ call it v5. I think that's better anyway. In this PR I've made the docs changes needed:
- renamed 4.2 to v5 everywhere
- polished the upgrade guide
- polished the release notes
- mentioned PRO for each PRO feature
- added installation steps for backpack/pro
- added explanation about open-core split to upgrade guide

### DIFF
--- a/5.x-dev/add-ons-community.md
+++ b/5.x-dev/add-ons-community.md
@@ -1,0 +1,19 @@
+# Community Add-ons
+
+---
+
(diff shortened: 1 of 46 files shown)


tabacitu:
You can see them here - https://backpackforlaravel.com/docs/5.x-dev

tabacitu:
After we _actually_ do the open-core split we'll just delete the 4.2.x docs in favor of v5.

PR #323, opened 2022-02-22
Remove "Custom" in namespaces

flartet:
I think it can ben confusing or can lead to copy-paste problems to use Admin/Operations namespace in one example and Admin/CustomOperations in another.
Furthermore, the generator does it in Admin/Operations namespace.
I might have missed something too in the doc logic but I don't think so.


tabacitu:
Totally agree. Thank you @flartet !

PR #342, opened 2022-05-10
update docs, remove unecessary setValidation call.

pxpm:
After we merged https://github.com/Laravel-Backpack/CRUD/pull/4220 this step is not necessary anymore. 

hurra \o


tabacitu:
Agreed - but we can't really merge this, because some people are using versions where this IS needed. So we gotta postpone this until we launch v6, I'm afraid 😔  Or instead of removing it, mention that that's needed only in 5.0, not starting with 5.1 - we can probably do that after we tag 5.1, yeah.

pxpm:
ping @tabacitu

PR #344, opened 2022-05-12
Update crud-operations.md

abejenaru:
Fix button view file reference (crud::buttons.button).
Replace usage of FontAwesome icons with LineAwesome.


tabacitu:
Hmm looks like this one slipped through the crack. I'm only seeing it now. Thank you @abejenaru 🙏

PR #346, opened 2022-05-19
Please add an example of how to add a crud field

firecentaur:
This is great that you explain that you can make a custom view - but very important documentation is missing here!!! Once you create a custom view, how on earth can I display a crud field?
Ie: I have a time field calledc "start_time" how do I display it in the custom view.  Please update the documentation to show an example!!!  Thanks!

### DIFF
--- a/5.x/crud-api.md
+++ b/5.x/crud-api.md
@@ -409,6 +409,10 @@ $this->crud->set('create.view', 'crud::yourfolder.yourview');
 $this->crud->set('create.view', 'resources.vendor.backpack.crud.yourfolder.yourview');
 ```
 
+This is great that you explain that you can make a custom view - but very important documentation is missing here!!! Once you create a custom view, how on earth can I display a crud field?
+Ie: I have a time field calledc "start_time" how do I display it in the custom view.  Please update the documentation to show an example!!!  Thanks!
+
+
 ### Content Class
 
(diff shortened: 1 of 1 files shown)


tabacitu:
Hey Paul,

I see you've sent 2 emails on this, posted on the blog, and posted here too. Let's keep this conversation as the final one please - best to keep this conversation public, so others who have this problem and search can find it.

This lack of documentation is intentional. While you certainly _can_ use the field blade files in your custom views, because you can just `@include('backpack::crud.fields.text', ['field' => [], 'entry' => $entry])`, we don't make it any easier at the moment because... it's not recommended to do so.

Fields are pretty coupled with the Create/Update forms where they're meant to be used. At the moment, it will be a bit tricky to use them outside (for example, you'll need to create two new blade stacks). And when we make changes to those files, we cannot take into account how/when you use them, so any change we make could be a breaking change for you, if you use them in a custom way.

I think it's a good idea for us to de-couple the fields from the forms. I do. It's something we plan to do in Backpack v6 - to turn them into Blade components, to make them easier to use like you intend. But until then, I really don't recommend you do that. If you do it... you're on your own 😀 We won't include docs for this because... we don't recommend this, sorry.

But then again, I've told people this before and they didn't listen soo... 🤷‍♂️😅

--

If you are set upon reusing the fields in a form, I heavily recommend you create a Backpack operation for it. Re-use as much as possible from the update/create forms, and customize those to your liking. Then you can re-use the fields, the same way the Create & Update operations do, because they're used the exact same way.

But in like custom forms. Custom dashboards. Custom screens. I don't recommend that, sorry.

Issue #347, opened 2022-05-25
Identifiable Attribute for Relationship Fields and AJAX


singeryo:
Hi !

When trying to implement a `select2_ajax` field I was puzzled as for why my identifiable attribute was not working properly, even though was following this guideline: https://backpackforlaravel.com/docs/5.x/crud-fields#optional-attributes-for-fields-containing-related-entries

I eventually realized I didn't know about this https://laravel.com/docs/8.x/eloquent-serialization#appending-values-to-json and it was exactly what i was missing.

Maybe add a note to the docs quickly explaining or reminding about this feature in Laravel ? Or maybe just add it the the `Category` example of that doc bit ?

Cheers

PR #348, opened 2022-06-04
mention appends for acessors in identifiable attribute

pxpm:
adds a mention to the use of `$appends` when using identifiableAttribute with acessors. 

Closes https://github.com/Laravel-Backpack/docs/issues/347

### DIFF
--- a/5.x/crud-fields.md
+++ b/5.x/crud-fields.md
@@ -262,7 +262,7 @@ class Category
     }
 }
 ```
-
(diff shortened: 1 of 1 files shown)


pxpm:
ping @tabacitu 

re-worded a bit and change to what I think is a more appropriate place.

tabacitu:
Muuch better place, I agree. Thanks, merged!

PR #351, opened 2022-06-10
Update base-how-to.md

realtebo:
Modified doc to fix an inconsistency
We must create the file in the same locations uses a few rows after


tabacitu:
Thanks @realtebo !

PR #352, opened 2022-06-10
Update crud-fields-javascript-api.md

realtebo:
Small fix after trying for the first  time, when we noticed that actual version generate a 404


tabacitu:
Ah, right! No `public`, thank you @realtebo 🙏

PR #360, opened 2022-07-08
add invokable stripRequest to docs

pxpm:
I replaced the ocurrences of `strippedRequest` with the invokable instead of closure.

image

I removed option 1 from `unchangeable-crud-field` because that option is not appropriate for that example. Added invokable as an option too.

### DIFF
--- a/5.x/crud-how-to.md
+++ b/5.x/crud-how-to.md
@@ -620,7 +620,7 @@ Please read the relationship [BelongsToMany](#belongstomany) documentation, ever
 
 You might want to add a new attribute to the Model that gets saved. Let's say you want to add an `updated_by` indicator to the Update operation, containing the ID of the user currently logged in (`backpack_user()->id`).
 
-Backpack uses the `strippedRequest` configuration to determine the fields that should be saved. **By default it will only save fields that have a corresponding CRUD field defined**
+Backpack uses the `strippedRequest` configuration to determine the fields that should be saved. **By default it will only save inputs that have a corresponding CRUD field defined.**
 
 **Option 1.** You can change the `strippedRequest` closure inside your `ProductCrudController::setup()`:
 ```php
--- a/5.x/crud-how-to.md
+++ b/5.x/crud-how-to.md
@@ -652,9 +652,9 @@ public function setupUpdateOperation()
     }
 ```
 
-**Option 3.** You can create an `invokable` class strippedRequest will use, acting like a closure, but can also used in config files. 
+**Option 3.** You can create an `invokable` class strippedRequest will use, acting like a closure.
 
-Create the invokable class:
+#####Create the invokable class:
 ```php
 <?php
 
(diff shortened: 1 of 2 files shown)


5.x/crud-how-to.md
```diff
@@ -615,15 +615,14 @@ Please read the relationship [BelongsToMany](#belongstomany) documentation, ever
...
-<a name="overwrite-a-method-on-the-crud-panel-object"></a>
-### Add an Uneditable Input inside Create or Update Operation
+<a name="add-non-editable-input-inside-create-or-update-operation-stripped-request"></a>
+### Add an Uneditable Input inside Create or Update Operation - Stripped Request
 
 You might want to add a new attribute to the Model that gets saved. Let's say you want to add an `updated_by` indicator to the Update operation, containing the ID of the user currently logged in (`backpack_user()->id`).
 
-**Option 1.** Sure, in your `ProductCrudController::setupUpdateOperation()` can do `CRUD::field('updated_by')->type('hidden')->value(backpack_user()->id);`, but because that hidden field is inside the HTML, it opens up the possibility that a malicious actor will edit the value of the input, in the browser.
-
+Backpack uses the `strippedRequest` configuration to determine the fields that should be saved. **By default it will only save fields that have a corresponding CRUD field defined**
```
tabacitu, line 623:
```suggestion
Backpack uses the `strippedRequest` configuration to determine the fields that should be saved. **By default it will only save inputs that have a corresponding CRUD field defined.**
```

5.x/upgrade-guide.md
```diff
@@ -116,24 +116,37 @@ No changes needed.
...
 ```diff
 -    'saveAllInputsExcept' => ['_token', '_method', 'http_referrer', 'current_tab', 'save_action'],
-+    'strippedRequest' => (function ($request) {
-+        return $request->except('_token', '_method', '_http_referrer', '_current_tab', '_save_action');
++    'strippedRequest' => '\App\Http\Requests\StripBackpackRequest',
 +    }),
 ```
-But you can also do a lot more, because you have the `$request` in that closure. See more info in PR #[3987](https://github.com/Laravel-Backpack/CRUD/pull/3987). In addition, please notice that **all hidden parameters are now prefixed by an underscore**. Starting with v5, if it starts with an underscore, you know it's not an actual database column.
-
+But you can also do a lot more, because you have the `$request` in that class. See examples here: https://backpackforlaravel.com/docs/5.x/crud-how-to#add-non-editable-input-inside-create-or-update-operation-stripped-request
```
tabacitu, line 145: We never want to show a full URL in the docs - that's ugly (1) and takes up too much space (2) which breaks the visual hierarchy - then the most obvious thing in that section is the big ass URL, which it shouldn't be, since that's just a sidenote. 

Always use hyperlinks please.

```suggestion
But you can also do a lot more, because you have the `$request` in that class. You can see [an example here]( https://backpackforlaravel.com/docs/5.x/crud-how-to#add-non-editable-input-inside-create-or-update-operation-stripped-request).
```

tabacitu (changes requested):

PR #361, opened 2022-07-13
Suggested method POST for when using `include_all_form_fields`

promatik:
Fix for https://github.com/Laravel-Backpack/CRUD/issues/4471.

Rephrased the "How to add a relationship field that depends on another field", to include a reference to the `relationship` field.

Also added a note to suggest the usage of `POST` method.

### DIFF
--- a/5.x/crud-how-to.md
+++ b/5.x/crud-how-to.md
@@ -199,9 +199,9 @@ function () { // if the filter is active (the GET parameter "draft" exits)
 This will make Backpack look for the ```resources/views/custom_filters/complex.blade.php```, and pick that up before anything else.
 
 <a name="how-to-add-a-select-that-depends-on-another-field"></a>
-### How to add a select2 field that depends on another field
+### How to add a relationship field that depends on another field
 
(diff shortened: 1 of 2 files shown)


5.x/crud-how-to.md
```diff
@@ -210,31 +210,35 @@ Say you want to show two selects:
...
+    'dependencies'            => ['category'], // when a dependency changes, this select2 is reset to null
+    // 'method'               => 'POST', // optional - HTTP method to use for the AJAX call (GET, POST)
 ]);
 ```
 
 **DIFFERENT HERE**: ```minimum_input_length```,  ```dependencies``` and ```include_all_form_fields```.
 
+Note: if you are going to use `include_all_form_fields` we recommend you to set the method to `POST`, and to properly setup that in your routes. Since all the fields in the form are going to be sent in the request, `POST` support more data.
+
 2. That second select points to routes that need to be registered:
```
tabacitu, line 242: Shouldn't we also use POST on the routes below, if we're to recommend using POST?

promatik: Makes total sense of course 👌

promatik: Done @tabacitu 👌
I also switched stuff in demo, from GET to POST; https://github.com/Laravel-Backpack/demo/pull/412.
Not 100% sure, but, I think we should right?

Issue #237, opened 2021-02-08
Some button methods are not documented


tabacitu:
Trying to answer [this SO question](https://stackoverflow.com/questions/66060885/reorder-line-buttons-in-backpack-laravel/66099152#66099152) I've realised that we don't actually tell people:
- how to add butons before/after other buttons
- how to move buttons before/after other buttons

At least not for the array syntax. So we should take another hard look at [the Buttons docs](https://backpackforlaravel.com/docs/4.1/crud-buttons#buttons-api) and Buttons methods and see which ones are missing. Then add them:
- inside [the Buttons API section](https://backpackforlaravel.com/docs/4.1/crud-buttons#buttons-api)
- inside the [API Cheat Sheet](https://backpackforlaravel.com/docs/4.1/crud-cheat-sheet#buttons)

At the very least, here are the ones I know for sure are not documented:
- `$this->crud->moveButton($target, $where, $destination)` - see [here](https://github.com/Laravel-Backpack/CRUD/blob/master/src/app/Library/CrudPanel/Traits/Buttons.php#L175-L209)
- `$this->crud->orderButtons(string $stack, array $order)` - see [here](https://github.com/Laravel-Backpack/CRUD/blob/master/src/app/Library/CrudPanel/Traits/Buttons.php#L22-L58)

PR #362, opened 2022-07-14
Added undocumented button methods

promatik:
Fix for https://github.com/Laravel-Backpack/docs/issues/237.

Added all undocumented button methods to `crud-buttons.md` and `crud-cheat-sheet.md` pages.

### DIFF
--- a/5.x/crud-buttons.md
+++ b/5.x/crud-buttons.md
@@ -34,7 +34,11 @@ Default buttons are invisible if an operation has been disabled. For example, yo
 Here are a few things you can call in your EntityCrudController's ```setupListOperation()``` method, to manipulate buttons:
 
 ```php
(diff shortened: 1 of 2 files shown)


tabacitu:
Excellent, thank you @promatik . Right now, this is a good bang-for-buck. But in the future, let's also consider adding a section with example usage [after defining the API](https://backpackforlaravel.com/docs/5.x/crud-buttons#buttons-api-1). Exactly the same, but with real parameters:

```diff

// add a button; possible types are: view, model_function
$this->crud->addButton('top', 'moderate', 'view', 'moderate', 'end');
```

Or maybe one example right after its definition would be best? So you can also see the variable name above? Yeah, probably. Nevertheless, what we have right now is fine for now.

Thanks @promatik !

PR #369, opened 2022-08-17
add week and month browser support notices

pxpm:
name says it all!

We are still working on possible solutions/workarounds, but ... it's not like 1+1=2!

This should be merged asap if everything ok, we later update with more info regarding our direction on that issue.

### DIFF
--- a/5.x/crud-fields.md
+++ b/5.x/crud-fields.md
@@ -511,6 +511,8 @@ Since not all browsers support this input type, if you are using [Backpack PRO](
     ],
 ]
 ```
+**Important**: you should be using a date/datetime column as database column type if using `date_picker`.
+
 <hr>
 
 <a name="number"></a>
--- a/5.x/crud-fields.md
+++ b/5.x/crud-fields.md
@@ -483,6 +483,9 @@ Include an `<input type="hidden">` in the form.
 <a name="month"></a>
 ### month
 
+**IMPORTANTE NOTICE**: This field type is not supported by all browsers. [View compatibility table](https://caniuse.com/mdn-html_elements_input_type_month)
+See `#workaround` section below for ideas on how to solve this.
+
 ```php
 [   // Month
     'name'  => 'month',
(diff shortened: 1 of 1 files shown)


5.x/crud-fields.md
```diff
@@ -495,6 +498,21 @@ Input preview:
 
 ![CRUD Field - month](https://backpackforlaravel.com/uploads/docs-4-2/fields/month.png)
 
+###`#workaround`
```
tabacitu, line 501:
```suggestion

```

5.x/crud-fields.md
```diff
@@ -483,6 +483,9 @@ Include an `<input type="hidden">` in the form.
 <a name="month"></a>
 ### month
 
+**IMPORTANTE NOTICE**: This field type is not supported by all browsers. [View compatibility table](https://caniuse.com/mdn-html_elements_input_type_month)
+See `#workaround` section below for ideas on how to solve this.
+
```
tabacitu, line 488:
```suggestion
Include an `<input type="month">` in the form.

**Important**: This input type is not supported by all browsers. [View compatibility table](https://caniuse.com/mdn-html_elements_input_type_month). See a workaround below.

```

5.x/crud-fields.md
```diff
@@ -1079,6 +1097,8 @@ Load a custom view in the form.
 <a name="week"></a>
 ### week
 
+**IMPORTANTE NOTICE**: This field type is not supported by all browsers. [View compatibility table](https://caniuse.com/mdn-html_elements_input_type_week)
+
```
tabacitu, line 1101:
```suggestion
Include an `<input type="week">` in the form.

**Important**: This input type is supported by most modern browsers, not all. [See compatibility table here](https://caniuse.com/mdn-html_elements_input_type_week).

```

5.x/crud-fields.md
tabacitu: Done this because it feel odd to start the section with CAREFUL, THIS DOESN'T WORK.

Also, rephrased so it's more positive - it DOES work, in most browsers. Doesn't in a few.

PR #370, opened 2022-08-30
show how to configure filter attributes

pxpm:
add example to docs on how to configure filter options.

### DIFF
--- a/5.x/crud-filters.md
+++ b/5.x/crud-filters.md
@@ -26,9 +26,8 @@ $this->crud->removeAllFilters();
 
 $this->crud->filters(); // gets all the filters
 
-// aditionally you should be able to change Filter attributes by appending `->methods` in your filter definition. Eg:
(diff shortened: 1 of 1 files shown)


5.x/crud-filters.md
```diff
@@ -25,6 +25,10 @@ $this->crud->removeFilter($name);
 $this->crud->removeAllFilters();
 
 $this->crud->filters(); // gets all the filters
+
+// aditionally you should be able to change Filter attributes by appending `->methods` in your filter definition. Eg:
+$this->crud->addFilter($options, $values, $filter_logic)->type('number');
+// this would set the filter attribute `type` to be equal to `number`.
```
tabacitu, line 31:
```suggestion
// aditionally you should be able to change Filter attributes by appending `->methods` in your filter definition. Eg - this would set the filter attribute `type` to be equal to `number`:
$this->crud->addFilter($options, $values, $filter_logic)->type('number');
```

tabacitu:
I... didn't know that's possible 😅🙏

PR #377, opened 2022-09-23
add docs for google map field

pxpm:


tabacitu:
Pedro let's please update the docs, so that "saving in two inputs" teaches devs to use Field Events, rather than hidden fields.

jorgetwgroup:
Look ready documentation.

Cheers

5.x/crud-fields.md
```diff
@@ -466,6 +466,73 @@ Input preview:
 
 <hr>
 
+<a name="google-map"></a>
+### google_map
```
tabacitu, line 470: I don't think this should be under FREE fields 👀 It's a PRO field, isn't it?

5.x/crud-fields.md
```diff
@@ -466,6 +466,73 @@ Input preview:
...
+                'lng' => $location->lng,
+                'full_address' => $location->formatted_address ?? ''
+            ];
+        }
+    );
+}
+
+```
+
+Input preview:
```
tabacitu, line 530: Let's upload an image for it here, inside the Github conversation. 
Then we can take that URL and introduce it in the markdown 😉

5.x/crud-fields.md
```diff
@@ -466,6 +466,73 @@ Input preview:
...
+// 
+protected function location(): \Illuminate\Database\Eloquent\Casts\Attribute
+{
+    return \Illuminate\Database\Eloquent\Casts\Attribute::make(
+        get: fn ($value, $attributes) => json_encode([
+            'lat' => $attributes['lat'], 
+            'lng' => $attributes['lng'], 
+            'formatted_address' => $attributes['full_address'] ?? ''
+            ],
+             JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_AMP | JSON_HEX_QUOT | JSON_THROW_ON_ERROR),
```
tabacitu, line 516: This looks weird to me on a separate line
```suggestion
            ], JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_AMP | JSON_HEX_QUOT | JSON_THROW_ON_ERROR),
```

5.x/crud-fields.md
```diff
@@ -466,6 +466,73 @@ Input preview:
...
+        set: function($value) {
+            $location = json_decode($value);
+            return [
+                'lat' => $location->lat,
+                'lng' => $location->lng,
+                'full_address' => $location->formatted_address ?? ''
+            ];
+        }
```
tabacitu, line 524: If the getter is using arrow functions, shouldn't the setter too? Let's choose one and stick with it.

5.x/crud-fields.md
```diff
@@ -466,6 +466,73 @@ Input preview:
 
 <hr>
 
+<a name="google-map"></a>
+### google_map
+
+Uses the Google Places API to generate a Map where user can navigate and select a position to store `Latitude` (**lat**), `Longitude` (**lng**) and the address string (**formatted_address**). By default it saves a string in dabase with `{lat: 123, lng: 456, formatted_address: 'Lisbon, Portugal'}`.
```
tabacitu, line 472: Sorry but no. This is NOT what the field does. Why are we writing from a technical perspective here? We should write from a USER perspective.

```suggestion
Shows a map and allows the user to navigate and select a position on that map (using the Google Places API). The field stores the latitude, longitude and the address string as a JSON in the database ( eg. `{lat: 123, lng: 456, formatted_address: 'Lisbon, Portugal'}`). If you want to save the info in separate db columns, continue reading below.
```

5.x/crud-fields.md
```diff
@@ -466,6 +466,73 @@ Input preview:
 
 <hr>
 
+<a name="google-map"></a>
+### google_map
+
+Uses the Google Places API to generate a Map where user can navigate and select a position to store `Latitude` (**lat**), `Longitude` (**lng**) and the address string (**formatted_address**). By default it saves a string in dabase with `{lat: 123, lng: 456, formatted_address: 'Lisbon, Portugal'}`. 
+**Note**: If you want two save in separate fields continue reading below.
```
tabacitu, line 473:
```suggestion
```

tabacitu (changes requested):

tabacitu:
Can I just say I absolutely LOVE the fact that @pxpm chose Lisbon for the screenshot location? Just like Jorge chose Chile for the `phone` field 😀 I love that these fields have a nod to its creator 😉 👏👏👏

PR #380, opened 2022-10-02
add notes about query orders and constrains

pxpm:
This are the notices we should add about https://github.com/Laravel-Backpack/CRUD/issues/4172


5.x/crud-operation-list-entries.md
```diff
@@ -31,6 +31,16 @@ For views, it uses:
...
+**NOTE**: If not added by developer, Backpack will append the `model key`(usually **id**) as an hidden column and order it by `DESC`. 
+You can modify this behavior by applying yourself an id order:
+```php
+protected function setupListOperation()
+{
+  $this->crud->orderBy('id', 'asc');
+}
+```
+Note that this is a `query constrain`, not a `table constrain`, so the reset button will not reset this order, it becomes the `base query` for that operation, so any filters etc you add on table will be applied AFTER the `query orders/constrains` you define on your setup operation.
+
```
tabacitu, line 43: Hmmm I think this is way too soon, at this point the user is just learning how everything works... not the right time to instruct about a quirk, if you ask me.

How above we add a subsection after "[Custom Query](https://backpackforlaravel.com/docs/5.x/crud-operation-list-entries#custom-query-1)" called "Custom Order" where we say everything we have to say about this?

karandatwani92:
Hello @pxpm, This branch is 27 commits behind master. I don't have authorization to merge new commits to this branch. Please pull those changes into this branch. This would give me more clarity, Afterward I make requested changes. Thanks

pxpm:
Thank you Karan. I will merge master here, but it should not change anything as this is a new addition to the docs. 

I just assigned you so that you get the opportunity to read a bit more of some "more deep" funtions of backpack. 

Everything is a learning process I like to think, I will merge master here and ping you back. 

Thanks 🙏

karandatwani92:
Hey @pxpm. I have made the changes as per @tabacitu direction.

Please confirm and merge the request. Thanks✌️

5.x/crud-operation-list-entries.md
```diff
@@ -206,6 +206,23 @@ $this->crud->limit();
 $this->crud->orderBy();
 // please note it's generally a good idea to use crud->orderBy() inside "if (!$this->crud->getRequest()->has('order')) {}"; that way, your custom order is applied ONLY IF the user hasn't forced another order (by clicking a column heading)
 ```
+**NOTE:** we call this constrains `query constrains`, and when you filter or search we call it `table constrains`. All the `query constrains` added on the setup operation cannot be reseted by `Reset Button`, and all the `table constrains` are applied on top of the `query constrains`.
```
tabacitu, line 209: 🚩 please double-check for grammar and typos.

I don't think it's worth giving names to this, it only complicates the explanation. How about:

```suggestion
**NOTE:** The query constraints added in the `setup()` method operation _cannot_ be reset by `Reset Button`. They are permanent for that CRUD, for all operation.
```

Does this transmit the same thing? I don't really know.

5.x/crud-operation-list-entries.md
```diff
@@ -206,6 +206,23 @@ $this->crud->limit();
...
+
+#### Custom Order
+
+<a name="custom-order"></a>
+
+List operation uses the `model key`(usually **id**) as default to order entries by `DESC`. You can modify this behavior by applying an order key:
```
tabacitu, line 215:
```suggestion
By default, the List operation gets sorted by the primary key (usually `id`), descending. You can modify this behaviour by defining your own ordering:
```

5.x/crud-operation-list-entries.md
```diff
@@ -206,6 +206,23 @@ $this->crud->limit();
...
+```php
+protected function setupListOperation()
+{
+    //change default order key if not passed by datatable
+    if (!$this->crud->getRequest()->has('order')){
+        $this->crud->orderBy('updated_at', 'desc');
+    }
+}
+```
+**NOTE**: This is a `query constrain`, not a `table constrain`, so the reset button will not reset this order if applied without `has('order')` check. It becomes the `base query` for that operation, so any filters etc you add on table will be applied AFTER the `query orders/constrains` you define on your setup operation.
```
tabacitu, line 225: I don't understand. Can you please rephrase so I understand? Why do we introduce these terms? Are they actually useful or do they just make it more difficult to understand?

```suggestion
**NOTE**: If you include the `has('order')` check, the reset button will be able to reset this order. Without this check, the order becomes permanent, so any filters etc you add on table will be applied _after_ the order you've defined.
```

tabacitu: Pedro this was your writing, right? 😅

tabacitu (changes requested):

5.x/crud-operation-list-entries.md
pxpm: Shame 👎

PR #384, opened 2022-10-18
Reflect migration to Line Awesome in examples

jordanella:
Instances of examples with icons that use fa and the fa- prefix replaced with those of la and la- respectively.


tabacitu:
Excellent - we somehow missed those, thank you @jordanella

PR #390, opened 2022-12-12
Fix options example

jorgetwgroup:
Fix options example on docs, because in "onlyCountries" list is not "cl" so if people copy/paste example will not work.

### DIFF
--- a/5.x/crud-fields.md
+++ b/5.x/crud-fields.md
@@ -1653,7 +1653,7 @@ Show a telephone number input. Lets the user choose the prefix using a flag from
     //  take note that options defined in `config` will override any default values from the field;
     'config' => [
         'onlyCountries' => ['bd', 'cl', 'in', 'lv', 'pt', 'ro'],
-        'initialCountry' => 'cl', // if restrict country with onlyCountries option need to be in list to work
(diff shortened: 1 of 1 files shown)


5.x/crud-fields.md
```diff
@@ -1652,8 +1652,8 @@ Show a telephone number input. Lets the user choose the prefix using a flag from
     // most options provided by intlTelInput.js are supported, you can try them out using the `config` attribute;
     //  take note that options defined in `config` will override any default values from the field;
     'config' => [
-        'onlyCountries' => ['pt', 'ro'],
-        'initialCountry' => 'cl',
+        'onlyCountries' => ['bd', 'cl', 'in', 'lv', 'pt', 'ro'],
+        'initialCountry' => 'cl', // if restrict country with onlyCountries option need to be in list to work
```
pxpm, line 1656:
```suggestion
        'initialCountry' => 'cl', // this needs to be in the allowed country list, either in `onlyCountries` or NOT in `excludeCountries`
```

pxpm (approved):
Just left a small text suggestion.

Thanks @jorgetwgroup I didn't notice my mistake 👍

PR #403, opened 2023-01-25
solves codeblock ending

Jodaille:
Hello,

merge seems to have removed end of PHP codeblock


tabacitu:
Ouch! Thanks a lot @Jodaille 🙏

Issue #406, opened 2023-02-04
404 not found index.md links


yarscript:
https://github.com/Laravel-Backpack/docs/blob/master/5.x/index.md
404 index.md links
<img width="832" alt="image" src="https://user-images.githubusercontent.com/26656665/216792770-3c62710c-8a1b-4346-a80a-1d282031ee8c.png">

yarscript:
https://github.com/Laravel-Backpack/docs/pull/407

tabacitu:
Fixed by https://github.com/Laravel-Backpack/docs/pull/407

PR #407, opened 2023-02-04
*.md doc files links fix

yarscript:
Please check if it's correct but it really looks like


yarscript:
https://github.com/Laravel-Backpack/docs/issues/406

tabacitu:
Hey @yarscript - thanks a lot for the PR! 🙏 The problem is... this fixes the index file to work in Github... but it will break it on our website - https://backpackforlaravel.com/docs 

I never thought of making this work on Github... I just _assumed_ most people will browse it on our website... and we prefer it that way, because the UX is better, we have... _search_ (😱). 

I would love to have a fix for BOTH github and our docs, but I don't really see a solution for that... and if I have to pick between the two... I'd prefer we keep it working on our website. So I'm going to have to close the PR, I'm afraid.

I wonder if there's a way to at least _mention_ to people that the index file is not working on Github though... To prevent any confusion like you had... or maybe we should add the index to the latest version on the README page too, so that people don't even get to `index.md`?! What do you think @yarscript ?

pxpm:
@tabacitu a permanent solution to this is to have an endpoint on the website responding to requests on `https://backpackforlaravel.com/docs/blob/****`.
- Grab the url.
- Parse `master => v5` etc
- Redirect to the doc page on website.

I don't foresee any troubles with this approach at the moment, and I think it would be a fair cost/benefit solution.

PR #424, opened 2023-05-04
[v6] add dropzone docs

pxpm:
docs for https://github.com/Laravel-Backpack/PRO/pull/125 including uploader and validation.

### DIFF
--- a/6.x/crud-fields.md
+++ b/6.x/crud-fields.md
@@ -1406,7 +1406,7 @@ Input preview:
 
 Show a [Dropzone JS Input](https://docs.dropzone.dev/).
 
-**Step 0.** Make sure the db column can hold the amount of text this field will have. For example, for MySQL, `VARCHAR(255)` might not be enough all the time (for 3+ files), so it's better to go with `TEXT`. Make sure you're using a big column type in your migration or db.
+**Step 0.** Make sure the model attribute can hold all the information needed. Ideally, your model should cast this attribute as `array` and your migration should make the db column either `TEXT` or `JSON`. Other db column types such as `VARCHAR(255)` might not be enough all the time (for 3+ files).
 
 **Step 1:** Add the `DropzoneOperation` to your `CrudController`
 
--- a/6.x/crud-fields.md
+++ b/6.x/crud-fields.md
@@ -1406,8 +1406,21 @@ Input preview:
 
 Show a [Dropzone JS Input](https://docs.dropzone.dev/).
 
+**Step 1:** Add the `DropzoneOperation` to your `CrudController`
+
 ```php
-[   
+class UserCrudController extends CrudController
+{
+    use \Backpack\CRUD\app\Http\Controllers\Operations\CreateOperation;
+    // ... other operations
+    use Backpack\Pro\Http\Controllers\Operations\DropzoneOperation;
+}
+```
+
+**Step 2:** Add the field in CrudController
+
+```php
+$this->crud->addField([   
     'name'  => 'photos',
     'label' => 'Photos',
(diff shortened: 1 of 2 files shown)


6.x/crud-fields.md
```diff
@@ -1401,6 +1401,43 @@ Input preview:
 
 <hr>
 
+<a name="dropzone"></a>
+### dropzone <span class="badge badge-pill badge-info">PRO</span>
```
tabacitu, line 1405: These instructions do not look complete to me. Doesn't this need to look more like this?

https://github.com/Laravel-Backpack/PRO/pull/125#issuecomment-1441730323

tabacitu: If I follow these instructions, will the field work? Or is there more information needed here, that I'm not given? Let's aim for a good experience.

6.x/crud-fields.md
```diff
@@ -1401,6 +1401,87 @@ Input preview:
...
+
+Show a [Dropzone JS Input](https://docs.dropzone.dev/).
+
+**Step 0.** Make sure the db column can hold the amount of text this field will have. For example, for MySQL, `VARCHAR(255)` might not be enough all the time (for 3+ files), so it's better to go with `TEXT`. Make sure you're using a big column type in your migration or db.
```
tabacitu, line 1409: Shouldn't we also tell people to cast this DB column to `array` in their Models?

6.x/crud-fields.md
```diff
@@ -1401,6 +1401,87 @@ Input preview:
...
+```php
+[   
+    'name'  => 'photos',
+    'label' => 'Photos',
+    'type'  => 'dropzone',
+    'withFiles' => true
+],
+```
+
+To know more about the `withFiles`, how it works and how to configure it, [ click here to read the documentation ](https://backpackforlaravel.com/docs/6.x/crud-uploaders).
```
tabacitu, line 1456: Let's tell about the simple option first, then tell people about the complex use case:

```suggestion
At this point you have the dropzone field showing up, and the ajax routes setup to upload/delete files, but the process is not complete. Your files are now only uploaded to the temporary folder, they need to be moved to the permanent location and their paths stored in the database. The easiest way to fix that is to add `withFiles => true` to your field definition, this will use the standard `AjaxUploader` that Backpack provides:

```php
[   
    'name'  => 'photos',
    'label' => 'Photos',
    'type'  => 'dropzone',
    'withFiles' => true
],
```

Alternatively, you can manually implement the saving process yourself using model events, mutators or any other solution that suits you. To know more about the `withFiles`, how it works and how to configure it, [read its documentation](https://backpackforlaravel.com/docs/6.x/crud-uploaders).
```

**Review by tabacitu — CHANGES_REQUESTED** [review]

File: 6.x/crud-fields.md (hunk 2 above)
Comment (pxpm) [reply 6.x/crud-fields.md:1409]: Only required if not using the `->withFiles()`.

We don't document how to do it without `->withFiles()`, just that it can be done using model events, mutators etc. 

So if developers are writing their own upload process they should be responsible/aware of what get's save in database and what gets retrieved. Casting is optional, they can do manual casting like we do on the uploader, so **I don't think there is the need to mention it here**. 

Let me know if you disagree and I can add it there as a note. 

Cheers

Comment (tabacitu) [reply 6.x/crud-fields.md:1409]:
```suggestion
**Step 0.** Make sure the model attribute can hold all the information needed. Ideally, your model should cast this attribute as `array` and your migration should make the db column either `TEXT` or `JSON`. Other db column types such as `VARCHAR(255)` might not be enough all the time (for 3+ files).
```

PR #426, opened 2023-05-05
temp folder docs

pxpm:
docs for https://github.com/Laravel-Backpack/CRUD/pull/4922


tabacitu (changes requested):
Completely agree with everything written here, excellent job 👏👏👏 But if we do what I mentione in https://github.com/Laravel-Backpack/CRUD/pull/4922 then we should also move this piece of docs to the Dropzone operation docs. And we don't need this extra docs page 🎉

PR #427, opened 2023-05-16
Tmp folder dropzone

pxpm:
Moved #426 to dropzone docs

### DIFF
--- a/6.x/crud-fields.md
+++ b/6.x/crud-fields.md
@@ -1415,7 +1415,7 @@ class UserCrudController extends CrudController
 {
     use \Backpack\CRUD\app\Http\Controllers\Operations\CreateOperation;
     // ... other operations
-    use Backpack\Pro\Http\Controllers\Operations\DropzoneOperation;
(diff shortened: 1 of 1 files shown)


6.x/crud-fields.md
```diff
@@ -1469,6 +1469,52 @@ use Backpack\Pro\Uploads\Validation\ValidDropzone;
...
+CRUD::setOperationSetting('purge_temporary_files_older_than', 72);
+```
+
+The first two options are self-explanatory. The third one is the number of hours after which temporary files are deleted and requires further atention and configuration.
+
+The process of file **deletion** does **not happen** automatically.
+In case of Dropzone we manually call the `PurgeTemporaryFolder` job whenever new files are uploaded so that we keep the dropzone temporary upload folder clean.
+
+To make the cleanup process more general, we created a command that you can run periodically, `backpack:purge-temporary-files`. 
+It accepts the following optional parameter:
```
tabacitu, line 1497:
```suggestion
It accepts the following optional parameters: 
```

tabacitu:
Excellent!

PR #428, opened 2023-05-19
Update crud-fields.md to include fake fields mutators

SilverComet89:


5.x/crud-fields.md
```diff
@@ -177,6 +177,14 @@ In this example, these 3 fields will show up in the create & update forms, the C
...
 If the ```store_in``` attribute wasn't used, they would have been stored in the ```extras``` column.
 
+Mutators will be against the ```store_in``` attribute, for the default ```extras``` value you would need a mutator such as:
+```php
+public function setExtrasAttribute($value)
+{
+	// Your code
+}
+```
+
```
pxpm, line 187:
```suggestion
If you need to perform any manipulation of the data before storing the fake fields, you should define a mutator for the column used to store the fake attributes, by default it's `extras`. It can be any other column you define in `store_in`. 
```php
// in YourModel.php
public function setExtrasAttribute($value)
{
	// Your code
}
```

```

pxpm:
Thanks @SilverComet89 🙏

PR #456, opened 2023-07-09
Update blade filename

peterhauke:
On the backpack installation I have just installed, the correct name of this file is:

` resources/views/vendor/backpack/ui/inc/menu_items.blade.php`

I am using Laravel Framework 10.14.1 and Backpack/crud 6.03 and backpack/theme-tabler 1.0.2

I am not aware if this is simply a mistake or the old filename.


pxpm:
It was indeed a mistake.

Thanks @peterhauke 🙏

PR #457, opened 2023-07-09
Make pronouns consistent

peterhauke:
It is acceptable to refer to an "admin" with "their" and in fact in the second part of the sentence "they" is used to refer to the "admin"


pxpm:
For consistency with the second part of the sentence I am going to merge this suggestion. 

Thanks @peterhauke 🙏

PR #458, opened 2023-07-09
Replace incorrect phrase "widget type" with "component"

peterhauke:
I think "widget type" is wrong here as this page is all about components. The Widget page mentions overriding "widget types".


pxpm:
thanks @peterhauke 🙏

PR #459, opened 2023-07-10
Correct/improve English and add example snippet

peterhauke:
The snippet makes it clear what the code should look like. 

Instead of "forget", the words "omit" or "don't specify" are clearer and better. I suggest "don't specify".

I thought of adding a screenshot, but I am not sure that it is necessary.

### DIFF
--- a/6.x/crud-fields.md
+++ b/6.x/crud-fields.md
@@ -172,7 +172,7 @@ You can now split your create/edit inputs into multiple tabs.
 
 ![CRUD Fields Tabs](https://backpackforlaravel.com/uploads/docs-4-0/operations/create_tabs.png)
 
-To use this feature, just specify the tab name for each of your fields. for example:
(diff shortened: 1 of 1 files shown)


6.x/crud-fields.md
```diff
@@ -172,14 +172,19 @@ You can now split your create/edit inputs into multiple tabs.
 
 ![CRUD Fields Tabs](https://backpackforlaravel.com/uploads/docs-4-0/operations/create_tabs.png)
 
-In order to use this feature, you just need to specify the tab name for each of your fields. Example:
+To use this feature, just specify the tab name for each of your fields. for example:
```
pxpm, line 175:
```suggestion
To use this feature, specify the tab name for each of your fields. For example:
```

pxpm:
Thanks again @peterhauke 🙏

PR #462, opened 2023-07-10
Fix typos

peterhauke:
I think "db" is not correct, it should either be "DB" or "database".


tabacitu:
Thanks Peter! We appreciate you submitting all of these 🙏

PR #464, opened 2023-07-11
Add an example to explanation of div widget and fix typos

peterhauke:
The description of the div widget was not immediately clear to me, so after experiments to see how the widget worked, I added an explanation for clarification.

### DIFF
--- a/6.x/base-widgets.md
+++ b/6.x/base-widgets.md
@@ -367,7 +367,10 @@ Allows you to include multiple widgets in the div attributes of your choice. For
 ]
 ```
 
-Anything you specify on this widget, other than ```type``` and ```content```, has to be a string, and will be considered an attribute of the "div" element.
+Anything you specify on this widget, other than ```type``` and ```content```, has to be a string, and will be considered an attribute of the "div" element. 
+For example, in the above snippet, the "div" element will end up with a ```class``` attribute with a value of ```row```, which is a string. 
+This value has to be a string of one of more class names, for example ```row border border-4 border-danger```.
+
 
 <hr>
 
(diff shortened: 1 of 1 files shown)


6.x/base-widgets.md
```diff
@@ -367,7 +367,10 @@ Allows you to include multiple widgets in the div attributes of your choice. For
 ]
 ```
 
-Anything you specify on this widget, other than ```type``` and ```content```, has to be a string, and will be considered an attribute of the "div" element.
+Anything you specify on this widget, other than ```type``` and ```content```, has to be a string, and will be considered an attribute of the "div" element. 
+For example, in the above snippet, the "div" element will end up with a ```class``` attribute with a value of ```row```, which is a string. 
+This value has to be a string of one of more class names, for example ```row border border-4 border-danger```.
```
pxpm, line 372:
```suggestion
For example, in the above snippet, the "div" element will end up with a ```class``` attribute with a value of ```row```, which is a string. 
This value has to be a string of one of more class names, for example ```row border border-4 border-danger```.
```

Maybe instead "verbosely" explaining this and having the need for two examples we:
- use two classes, `'class' => 'row my-custom-widget-class',`
- add one more attribute, `'custom-attribute' => 'my-custom-value'`
- show the generated output `<div custom-attribute="my-custom-value" class="row my-custom-widget-class">`

Do you think it will be better like I suggested above ?

peterhauke: Yes, that is much better. I had thought of adding the following to the widget:

'style' => 'border: solid 3px  red'

But your three suggestions together should be much clearer.

pxpm:
Thanks again @peterhauke

Issue #468, opened 2023-07-06
[Bug] Docs Base > How To page needs updating to reflect the new file/directory structure


fronbow:
# Bug report

### What I did

Just going through the installation to check the difference between this theme and the default tabler theme

### What I expected to happen

I expected the file structure to be like it is in the docs!

### What happened

The file/directory structure has changed

### Context to replicate the bug - Backpack, Laravel, PHP, DB version

When I run ```php artisan backpack:version``` the output is:
### PHP VERSION:
PHP 8.1.12-1ubuntu4 (cli) (built: Feb 22 2023 19:48:21) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.12, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.12-1ubuntu4, Copyright (c), by Zend Technologies
    with Xdebug v3.2.0, Copyright (c) 2002-2022, by Derick Rethans

### LARAVEL VERSION:
10.14.1.0

### BACKPACK PACKAGE VERSIONS:
backpack/basset: 1.0.0
backpack/crud: 6.0.1
backpack/generators: v4.0.0
backpack/permissionmanager: 7.0.0
backpack/pro: 2.0.5
backpack/theme-coreuiv4: 1.0.0
backpack/theme-tabler: 1.0.1

Cheers

tabacitu:
Ouch! Indeed that page needs A LOT of love. Thanks @fronbow , on it!

tabacitu:
I've just updated the HowTo page 🎉 In most cases, I've pointed the user to the theme docs. 

Regarding the folder structure, each theme can now have its own folder structure... so we can't really teach anybody about the folder structure in the docs. So we point them to look inside the theme they are using.

PR #469, opened 2023-07-12
Fix minor typos, etc.

peterhauke:
I think the brackets were in the wrong place. Or possibly, brackets are not really necessary in this context.

"plain old" is not hyphenated.

### DIFF
--- a/6.x/base-about.md
+++ b/6.x/base-about.md
@@ -34,7 +34,7 @@ This  file is used to show the contents of the menu. It's been published there s
 
 **2) Middleware to ```app/Http/Middleware/CheckIfAdmin.php```**
 
-This middleware is used to test if users have access to admin panel pages. You can (and should customize it) if you have both users and admins in your app.
+This middleware is used to test if users have access to admin panel pages. You can (and should) customize it if you have both users and admins in your app.
 
 **3) Route file to ```routes/backpack/custom.php```**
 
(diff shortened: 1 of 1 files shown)


peterhauke:
On line 94, I did not really understand what "Or that you plan to use it for both users & admins." means and how exactly it connects with the previous sentence. So I did not make any changes to that sentence.

pxpm:
Thanks @peterhauke 

I will refrain from merging this one right now because @tabacitu is going to work on this page soon as per https://github.com/Laravel-Backpack/docs/issues/468

For example the `error views` text, we no longer publish error views anymore, so we should remove that, and add instructions on how they work and how they can be customized. 

Thanks 🙏

tabacitu:
Yup, on it! Hopefully later today 🤞

peterhauke:
Thanks @tabacitu , @pxpm 

If you want to assign me a page to look at, then please do so. I quite often have time to read through and correct one page like this a day.

Normally, I am just looking:
- for mistakes;
- to eliminate inconsistencies; and
- to clarify ambiguous text.
 
I try not to change the writing style too much.

tabacitu:
That would be very helpful, yes, thank you. I'll take you up on it 😉 Right now what this page needs... is for us to change to logic of it first. There are things here that were true in v5 but no longer true in v6.

PR #478, opened 2023-07-26
Couple of code examples updated and their formatting made consistent

tomsb:


pxpm:
Hey @tomsb thanks for the PR 🙏 

The only thing I think we should merge is the fix in the `join` example by changing `select` (the name of the column we use in demo 🙃 by `category_id`, what makes sense in a real project. 

Can you please clean up this PR a leave only that change ? 

We can discuss the other later, but I really don't think they are worth poluting the docs with. 

Once again, thank you very much 🙏 

Cheers

tomsb:
Removed the polluting stuff, but the other changes look good (some formatting and bug fixes)

PR #500, opened 2023-08-21
Fix file reference in 6.x crud-columns

JeremieConstant:
The documentation stated that the "default_date_format" and "default_datetime_format" are located in /config/backpack/base.php when in reality they are in /config/backpack/ui.php


pxpm:
Thanks @JeremieConstant 🙏 

Much appreciated!

PR #503, opened 2023-08-29
add email validation

pxpm:
Docs for https://github.com/Laravel-Backpack/CRUD/pull/5301

### DIFF
--- a/6.x/base-how-to.md
+++ b/6.x/base-how-to.md
@@ -418,7 +418,7 @@ This will make the registration process pick up a view you can create, in ```res
 
 In Backpack CRUD 6.2 we introduced the ability to require email verification when accessing Backpack routes. To enable this feature please do the following: 
 
-**Step 1** - Ensure your Authenticable model (usually `App\Models\User`) implements the `Illuminate\Contracts\Auth\MustVerifyEmail` contract.
+**Step 1** - Make sure your user model (usually `App\Models\User`) implements the `Illuminate\Contracts\Auth\MustVerifyEmail` contract. [More info](https://laravel.com/docs/10.x/verification#model-preparation).
 
 ```php
 <?php
--- a/6.x/base-how-to.md
+++ b/6.x/base-how-to.md
@@ -438,11 +438,9 @@ class User extends Authenticatable implements MustVerifyEmail
 ```
 More info in: [Preparing model for email verification](https://laravel.com/docs/10.x/verification#model-preparation)
 
-**Step 2** - Make sure your Authenticable model table has a `email_verified_at`timestamp column.
+**Step 2** - Make sure your user model table has a `email_verified_at` column (timestamp). New Laravel installations already have it, but if you upgraded from L8/L9 it's possible that the `email_verified_at` column is missing.
 
-New Laravel instalations already bring this column, but if you came from earlier versions it's possible that `email_verified_at` column is missing. 
-
-You can quickly create a new migration that add that column with the following code: (please adjust table name if different from default). 
+You can quickly create a new migration using `php artisan make:migration add_email_verified_at_column_to_users --table=users`, then use the code below:
 
 ```php
 <?php
@@ -475,7 +473,7 @@ return new class extends Migration
 };
 
 ```
-More info in: [Preparing database for email verification](https://laravel.com/docs/10.x/verification#database-preparation)
+Then run `php artisan migrate`. [More info](https://laravel.com/docs/10.x/verification#database-preparation).More info in: [Preparing database for email verification](https://laravel.com/docs/10.x/verification#database-preparation)
 
 **Step 3** - Make sure you have the `verified` and `signed` middleware alias in your `App\Http\Kernel.php` file. New Laravel 10 installations already have them, but if you came from earlier versions it's possible that they are missing.
 
--- a/6.x/base-how-to.md
+++ b/6.x/base-how-to.md
@@ -413,3 +413,59 @@ Add whatever validation rules & inputs you want, in addition to name and passwor
 ```
 This will make the registration process pick up a view you can create, in ```resources/views/vendor/backpack/ui/auth/register.blade.php```. You can copy-paste the original view, and modify as you please. Including adding your own custom inputs.
 
+<a name="enable-email-verification-in-backpack-routes"></a>
+### Enable email verification in Backpack routes
+
+In Backpack CRUD 6.2 we introduced the ability to require email verification when accessing Backpack routes.
+
+To enable this feature please do the following: 
+
+**Step 1** - Ensure your Authenticable model (usually `App\Models\User`) implements the `Illuminate\Contracts\Auth\MustVerifyEmail` contract.
+
+```php
+<?php
+ 
+namespace App\Models;
+ 
+use Illuminate\Contracts\Auth\MustVerifyEmail;
+use Illuminate\Foundation\Auth\User as Authenticatable;
+use Illuminate\Notifications\Notifiable;
+ 
+class User extends Authenticatable implements MustVerifyEmail
+{
+    use Notifiable;
+ 
+    // ...
+}
+```
+More info in: [Preparing model for email verification](https://laravel.com/docs/10.x/verification#model-preparation)
+
+**Step 2** - Make sure your Authenticable model table has a `email_verified_at`timestamp column.
+
+New Laravel instalations already bring this column, but if you came from earlier versions it's possible that `email_verified_at` column is missing. 
+
+You can quickly create a new migration that add that column with the following code: (please adjust table name if different from default). 
+
+```php
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->addColumn('timestamp', 'email_verified_at', ['nullable' => true])->after('email');
+        });
+    }
+};
+```
+More info in: [Preparing database for email verification](https://laravel.com/docs/10.x/verification#database-preparation)
+
+**Step 3** - Enable the functionality by changing `setup_email_validation_routes` in `config/backpack/base.php` to `true`.
+
+If you don't have this config there, it's a good time to add it. 
+
+
--- a/6.x/base-how-to.md
+++ b/6.x/base-how-to.md
@@ -476,8 +476,7 @@ protected $middlewareAliases = [
     ];
 ```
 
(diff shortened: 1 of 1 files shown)


tabacitu (approved):

6.x/base-how-to.md
```diff
@@ -413,3 +413,71 @@ Add whatever validation rules & inputs you want, in addition to name and passwor
...
 This will make the registration process pick up a view you can create, in ```resources/views/vendor/backpack/ui/auth/register.blade.php```. You can copy-paste the original view, and modify as you please. Including adding your own custom inputs.
 
+<a name="enable-email-verification-in-backpack-routes"></a>
+### Enable email verification in Backpack routes
+
+In Backpack CRUD 6.2 we introduced the ability to require email verification when accessing Backpack routes.
+
+To enable this feature please do the following: 
+
+**Step 1** - Ensure your Authenticable model (usually `App\Models\User`) implements the `Illuminate\Contracts\Auth\MustVerifyEmail` contract.
```
tabacitu, line 423:
```suggestion
**Step 1** - Make sure your user model (usually `App\Models\User`) implements the `Illuminate\Contracts\Auth\MustVerifyEmail` contract. [More info](https://laravel.com/docs/10.x/verification#model-preparation).
```

6.x/base-how-to.md
```diff
@@ -413,3 +413,71 @@ Add whatever validation rules & inputs you want, in addition to name and passwor
...
+    // ...
+}
+```
+More info in: [Preparing model for email verification](https://laravel.com/docs/10.x/verification#model-preparation)
+
+**Step 2** - Make sure your Authenticable model table has a `email_verified_at`timestamp column.
+
+New Laravel instalations already bring this column, but if you came from earlier versions it's possible that `email_verified_at` column is missing. 
+
+You can quickly create a new migration that add that column with the following code: (please adjust table name if different from default).
```
tabacitu, line 447: These are not three ideas, but one. So it doesn't need three paragraphs, but one. Prettier but also more readable.

```suggestion
**Step 2** - Make sure your user model table has a `email_verified_at` column (timestamp). New Laravel installations already have it, but if you upgraded from L8/L9 it's possible that the `email_verified_at` column is missing. You can quickly create a new migration to add that column (example below) then run `php artisan migrate` to run it. [More info(https://laravel.com/docs/10.x/verification#database-preparation).
```

6.x/base-how-to.md
```diff
@@ -413,3 +413,71 @@ Add whatever validation rules & inputs you want, in addition to name and passwor
...
+return new class() extends Migration {
+    public function up()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->addColumn('timestamp', 'email_verified_at', ['nullable' => true])->after('email');
+        });
+    }
+};
+```
+More info in: [Preparing database for email verification](https://laravel.com/docs/10.x/verification#database-preparation)
```
tabacitu, line 465: Moved it before the code snippet. That way the article flows better - Step text, Step code, Step text, Step code. Easier on the eyes, easier to read.
 
```suggestion
```

6.x/base-how-to.md
```diff
@@ -413,3 +413,71 @@ Add whatever validation rules & inputs you want, in addition to name and passwor
...
+use Illuminate\Notifications\Notifiable;
+ 
+class User extends Authenticatable implements MustVerifyEmail
+{
+    use Notifiable;
+ 
+    // ...
+}
+```
+More info in: [Preparing model for email verification](https://laravel.com/docs/10.x/verification#model-preparation)
```
tabacitu, line 441: Moved above. Same reason as below - text flows better. I agree this would have been the perfect place to put an "extras" link normally. But since this text is pretty long, it's just simpler to have all text in one place, I think.

```suggestion
```

6.x/base-how-to.md
```diff
@@ -413,3 +413,71 @@ Add whatever validation rules & inputs you want, in addition to name and passwor
...
+        'verified' => \Illuminate\Auth\Middleware\EnsureEmailIsVerified::class,
+        // if you don't have the VaidateSignature middleware you can copy it from here: 
+        // https://github.com/laravel/laravel/blob/10.x/app/Http/Middleware/ValidateSignature.php
+        'signed' => \App\Http\Middleware\ValidateSignature::class,
+    ];
+```
+
+**Step 4** - Enable the functionality by changing `setup_email_validation_routes` in `config/backpack/base.php` to `true`.
+
+If you don't have this config key there, now it's a good time to add it.
```
tabacitu, line 481: - Reordered the instruction a bit, to have the scope first (the config), then narrow down to what to do. Makes it easier to read.
- Merged the two sentences into one paragraph. It's only one idea, not two, so why two paragraphs?

```suggestion
**Step 4** - Enable the functionality in `config/backpack/base.php` by changing `setup_email_validation_routes` to `true`. If you don't have this config key there, now is a good time to add it. 
```

6.x/base-how-to.md
```diff
@@ -413,3 +413,71 @@ Add whatever validation rules & inputs you want, in addition to name and passwor
 ```
 This will make the registration process pick up a view you can create, in ```resources/views/vendor/backpack/ui/auth/register.blade.php```. You can copy-paste the original view, and modify as you please. Including adding your own custom inputs.
 
+<a name="enable-email-verification-in-backpack-routes"></a>
+### Enable email verification in Backpack routes
+
+In Backpack CRUD 6.2 we introduced the ability to require email verification when accessing Backpack routes.
+
+To enable this feature please do the following:
```
tabacitu, line 421: Too many paragraphs. It's one idea, so it doesn't need two paragraphs. We should avoid having short but many paragraphs - it does not read easily, and it does not look good. The rule of thumb is one-idea-one-paragraph. Sometimes it's not possible and we can split one idea into multiple paragraphs. But only when each paragraph is "long enough". If paragraphs are short and the same idea, they need squishing 😀

```suggestion
In Backpack CRUD 6.2 we introduced the ability to require email verification when accessing Backpack routes. To enable this feature please do the following: 
```

tabacitu (changes requested):
I love it! Excellent docs here @pxpm . My only real note here is regarding paragraphing. So I'm not going to merge myself, I'll let you do it to see WHY I suggest these things.

6.x/base-how-to.md
```diff
@@ -413,3 +413,71 @@ Add whatever validation rules & inputs you want, in addition to name and passwor
...
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    public function up()
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->addColumn('timestamp', 'email_verified_at', ['nullable' => true])->after('email');
+        });
+    }
+};
```
tabacitu, line 463:
```suggestion
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

return new class extends Migration
{
    /**
     * Run the migrations.
     */
    public function up(): void
    {
        Schema::table('users', function (Blueprint $table) {
            $table->addColumn('timestamp', 'email_verified_at', ['nullable' => true])->after('email');
        });
    }

    /**
     * Reverse the migrations.
     */
    public function down(): void
    {
        Schema::table('users', function (Blueprint $table) {
            $table->dropColumn(['email_verified_at']);
        });
    }
};

```

6.x/base-how-to.md
pxpm: image

I agree with all the rest tho.

6.x/base-how-to.md
pxpm: I look at your text now and I only see: 
image

It can be improved I agree, with your wording more than mine, but maybe separate the "fix" for those who need it ?f
```suggestion
**Step 2** - Make sure your user model table has a `email_verified_at` column (timestamp). New Laravel installations already have it, but if you upgraded from L8/L9 it's possible that the `email_verified_at` column is missing.

You can quickly create a new migration using `php artisan make:migration add_email_verified_at_column_to_users --table=users`, use the code below, then run `php artisan migrate`. [More info](https://laravel.com/docs/10.x/verification#database-preparation).
```
I still don't like that I have to read the code below, then remember that I need to run artisan migrate. Probably will re-read.

While if they are in order, you just follow:
- use the code below
**code**
- run migrate
- advance step

But I agree with "page flow", not sure it's the best for a step by step tutorial (basically it's what this is).

6.x/base-how-to.md
tabacitu: Exactly. We're not formal here. We shouldn't be. We're developers talking to developers.

6.x/base-how-to.md
pxpm: As you see I tried to make it small as possible even removed the comment, that page is already HUGE. 

image

We may need `how-to-look-and-feel` and `how-to-autentication`

And I noticed that I need to add this in the top index too 👍

6.x/base-how-to.md
tabacitu: Hmm good point. That is a separate idea here, and it's long, so a new paragraph is needed. Plus that sentence is waaay to long and complex, so maybe we split it up into multiple sentences - easier to read that way.

How about this?

```suggestion
**Step 2** - Make sure your user model table has a `email_verified_at` column (timestamp). New Laravel installations already have it, but if you upgraded from L8/L9 it's possible that the `email_verified_at` column is missing.

You can quickly create a new migration using `php artisan make:migration add_email_verified_at_column_to_users --table=users`, then use the code below:
```

6.x/base-how-to.md
tabacitu: Ok let's do the migrate step here:

```suggestion
Then run `php artisan migrate`. [More info](https://laravel.com/docs/10.x/verification#database-preparation).More info in: [Preparing database for email verification](https://laravel.com/docs/10.x/verification#database-preparation)
```

6.x/base-how-to.md
tabacitu: I don't understand. Please rephrase?

tabacitu:
Xnapper-2023-08-31-09 59 10

6.x/base-how-to.md
```diff
@@ -413,3 +413,80 @@ Add whatever validation rules & inputs you want, in addition to name and passwor
...
+    public function down(): void
+    {
+        Schema::table('users', function (Blueprint $table) {
+            $table->dropColumn(['email_verified_at']);
+        });
+    }
+};
+
+```
+Then run `php artisan migrate`. [More info](https://laravel.com/docs/10.x/verification#database-preparation).More info in: [Preparing database for email verification](https://laravel.com/docs/10.x/verification#database-preparation)
```
pxpm, line 475:
```suggestion
Then run `php artisan migrate`. [More info](https://laravel.com/docs/10.x/verification#database-preparation).
```

PR #506, opened 2023-08-31
Add a reference to setFromDb()

peterhauke:
There does not appear to be any reference to this function in the documentation, however it appears in the setupListOperation() and setupCreateOperation() when creating a CrudController, so I added a description (adapted from setFromDb() in AutoSet.php)

### DIFF
--- a/6.x/crud-api.md
+++ b/6.x/crud-api.md
@@ -65,6 +65,12 @@ $this->crud->setColumns();
 // sets the columns you want in the table view, either as array of column names, or multidimensional array with all columns detailed with their types
 ```
 
+- **setFromDb()** - use the public columns in the database table;
+```php
+$this->crud->setFromDb();
+// for a simple CRUD Panel, there should be no need to add/define the fields/columns. The public columns in the database will be converted to be fields
+```
+
 - **Chained - beforeColumn()** - insert current column _before_ the given column
 ```php
 // ------ REORDER COLUMNS


karandatwani92:
Hey, @tabacitu `setFromDb()` is missing in the CRUD **API list**.

Please review and merge.

Thanks @peterhauke

tabacitu:
Thanks for the note @peterhauke !

Actually @karandatwani92 we don't want people using `setFromDb()` - it's deprecated. So instead, let's remove any mentions of it from the docs, please 🙏

Why? Because `setFromDb()` isn't good-enough in the first place. It doesn't properly guess column&field types. And because of a way it was constructed, we can never _ever_ make updates to it. Any updates we do make... are a breaking change to everybody who's using it. That's why we've deprecated it.

PS. I don't know if it already has a `@deprecated` note on the docblock, in the code. If it does not, please do that too @karandatwani92

peterhauke:
I had a look at `setFromDb()` in [autoset.php](https://github.com/Laravel-Backpack/CRUD/blob/4cd8e5d909c77a4d7ce2bfbc7eac46490589ad20/src/app/Library/CrudPanel/Traits/AutoSet.php#L13) and there does not appear to be a deprecated note there.

I had a look at `setupListOperation()` in [crud-controller.stub](https://github.com/Laravel-Backpack/Generators/blob/08a61a0b11b7c657189ffa6bcd45d0b9c28450ef/src/Console/stubs/crud-controller.stub#L42) and there is no mention of it being deprecated there either. 

`setFromDb()` is currently being included in the code when a new crud controller is created.

Sorry for asking, but is there another function to use instead of `setFromDb()` ?

karandatwani92:
Yes, Not appropriate to `@deprecate` as it's in [crud-controller.stub](https://github.com/Laravel-Backpack/Generators/blob/08a61a0b11b7c657189ffa6bcd45d0b9c28450ef/src/Console/stubs/crud-controller.stub#L42) .

 @tabacitu please give directions.

promatik:
@karandatwani92 it's true the `setFromDb` is in the stubs, it should be removed from there, anyway, in the between we can keep it as an example, but developers will be warned that it is deprecated.
I'll close this one, let's move conversation here; https://github.com/Laravel-Backpack/CRUD/pull/5419

PR #507, opened 2023-09-01
Add example based on column available in FREE package

peterhauke:
The table type column is a PRO feature and so the table type example does nothing and will confuse anybody testing Backpack before committing to purchase. I choose the custom_html type because it did not need to access a database.


tabacitu:
I agree - thanks for the PR @peterhauke !

PR #509, opened 2023-09-06
Update crud-operation-update.md

undersound:
The second sentence seems to wanna give an answer to the question asked in the first sentence. Using the word "then" does not really make sense here.

### DIFF
--- a/6.x/crud-operation-update.md
+++ b/6.x/crud-operation-update.md
@@ -155,7 +155,7 @@ protected function setupUpdateOperation()
 
 This works particularly well for bigger models, because you can mention a lot of rules, messages and attributes in your `FormRequest` and it will not increase the size of your `CrudController`.
 
-**Differences between the Create and Update validations?** They create a separate request file for each operation and instruct your EntityCrudController to use those files:
+**Do you need different Create and Update validations?** Then create a separate request file for each operation and instruct your EntityCrudController to use those files:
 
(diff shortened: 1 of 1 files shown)


promatik:
Hi @undersound, the "then" makes sense there...

For the question; "Differences between the Create and Update validations?"

Meaning;
_Do you need to differentiate Create and Update validations?_ **then** do this...

But I agree this is a little bit confusing, specially because of the first sentence.
I suggest this change.

6.x/crud-operation-update.md
```diff
@@ -155,7 +155,7 @@ protected function setupUpdateOperation()
 
 This works particularly well for bigger models, because you can mention a lot of rules, messages and attributes in your `FormRequest` and it will not increase the size of your `CrudController`.
 
-**Differences between the Create and Update validations?** Then create a separate request file for each operation and instruct your EntityCrudController to use those files:
+**Differences between the Create and Update validations?** They create a separate request file for each operation and instruct your EntityCrudController to use those files:
```
promatik, line 158:
```suggestion
**Do you need different Create and Update validations?** Then create a separate request file for each operation and instruct your EntityCrudController to use those files:
```

undersound:
_Do you need to differentiate Create and Update validations? then do this..._

Ah I see, that makes it more clear indeed.

PR #514, opened 2023-09-29
Add livewire widget

pxpm:


6.x/base-widgets.md
```diff
@@ -414,6 +414,64 @@ Widget Preview:
...
+```blade
+<!-- livewire/hello-world.blade.php -->
+<div>
+    Hello {{ $name }}
+</div>
+```
+
+```php
+// add the widget to the page
+Widget::add()->type('livewire')->content('hello-world')->parameters(['name' => 'John Doe'])->wrapperClass('col-md-12 text-center');
```
tabacitu, line 466: Could we use “component” instead of “content”? Would it be smart or atypical?

6.x/base-widgets.md
```diff
@@ -414,6 +414,64 @@ Widget Preview:
 
 <hr>
 
+<a name="livewire"></a>
+### Livewire
+
+Add a Livewire component to a page. If you haven't created your component yet, head to [Livewire documentation](https://livewire.laravel.com/docs/components) and create the component you want to use.
+
+**Important Note**: This widget automatically load the Livewire Assets (scripts and styles) for you. If you are using Livewire in the same page as this widget and is already loading the assets, you can disable the automatic loading by setting the ```livewireAssets``` attribute to ```false```.
```
tabacitu, line 422: Shouldn’t we default to false? Assume people are using Livewire v3?

tabacitu (changes requested):

6.x/base-widgets.md
pxpm: "content" is on every widget from divs to cards, it seem'd to me that giving the "content" of the "livewire" widget made sense too. I like it more than creating a different key. 

can always revert, or use both as aliases 🙏

6.x/base-widgets.md
pxpm: that's the tricky question and why I pinged you. 

It makes sense to be false by default especially for Lw3, I just wanted to make sure we are not giving hard time for other users.

tabacitu: I think that’s ok, as long as it’s documented

6.x/base-widgets.md
tabacitu: Ok let’s keep content

6.x/base-widgets.md
tabacitu: I mean yes let’s default to false

PR #519, opened 2023-10-20
docs for columns `linkTo` helper

pxpm:
docs for https://github.com/Laravel-Backpack/CRUD/pull/5317

### DIFF
--- a/6.x/crud-columns.md
+++ b/6.x/crud-columns.md
@@ -1541,7 +1541,44 @@ $this->crud->addColumn([
 
 Sometimes the text that the column echoes is not enough. You want to add interactivity to it, by adding a link to that column. Or you want to show the value in a green/yellow/red badge so it stands out. You can do both of that - with the ```wrapper``` attribute, which most columns support.
 
-For example, you can wrap the text in an anchor element, to point to that Article's Show operation:
+The easiest and most straightfoward way to get started with it, is using our column helper `linkTo($routeName, $target = null)`. This helper will set the `href` attribute automatically for you, **including related entries**.
+
+```php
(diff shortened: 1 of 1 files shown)


6.x/crud-columns.md
```diff
@@ -1541,10 +1541,48 @@ $this->crud->addColumn([
...
+
+// array syntax is also supported
+
+$this->crud->column([
+    'name' => 'category',
+    'linkTo' => 'category.show',
+
+    // alternatively with target configuration
+    'linkTo' => [
+        'name' => 'category.show',
```
tabacitu, line 1573:
```suggestion
        'routeName' => 'category.show',
```

tabacitu:
Pedro I like the text you wrote, but I didn't like that it was crowded inside the `wrapper` docs. I moved it outside in its own section, I think that will be clearer... and will give us the option of linking to this feature in particular.

6.x/crud-columns.md
```diff
@@ -1589,6 +1587,49 @@ $this->crud->addColumn([
...
+
+// array syntax is also supported
+
+$this->crud->column([
+    'name' => 'category',
+    'linkTo' => 'category.show',
+
+    // alternatively with target configuration
+    'linkTo' => [
+        'nameName' => 'category.show',
```
tabacitu, line 1624: Heads-up, the docs said `name` here but I discovered what works is actually `routeName`. So I used that. I don't love it to be honest, I would rather have `route` here... what do you think?

tabacitu (approved):

tabacitu:
This is excellent! Ready to go from my point of view, take it away @pxpm / @promatik !

PR #521, opened 2023-11-03
Update instruction on how to customize reg form

phpfour:
The previous instruction on how to override the registration form does not work. This is a working version.


phpfour:
Hey @tabacitu, I'm not sure if this PR still makes sense...I am unable to assign it to you, so just unassigning myself 🙏🏼

PR #529, opened 2023-12-08
add dcs for error codes when installing paid add-ons

tabacitu:
While replying to someone on email, I realized we don't have docs on the most common errors, when installing paid packages. This adds those docs.


tabacitu:
Merging this. Team, feel free to make changes if I didn't phrase something right.

PR #536, opened 2024-01-08
Add aria-label example to Crud Fields documentation (for accessibility)

tibbsa:
In circumstances where the default Backpack field implementation is not accessible to screen readers and other assistive technologies (and in the absence of any movement on the related bug report (https://github.com/Laravel-Backpack/community-forum/issues/206), I propose to at least actively document and highlight the way to work around this for most of the common field types.   

Note, this would apply to earlier versions as well, but in case there is some need to change the wording I have currently only put this into the 6.x branch.

This, of course, does not ensure accessibility of all field types, but it is better than having no access at all to even basic fields.

### DIFF
--- a/6.x/crud-fields.md
+++ b/6.x/crud-fields.md
@@ -112,6 +112,37 @@ These will help you:
 - **attributes** - change or add actual HTML attributes of the input (ex: readonly, disabled, class, placeholder, etc);
 - **wrapper** - change or add actual HTML attributes to the div that contains the input;
 
+<a name="optional-field-attributes-for-accessibility"></a>
+#### Optional but Recommended - Field Attributes for Accessibility
+
+By default, field labels are not directly associated with input fields. To improve the accessibility of CRUD fields for screen readers and other assistive technologies (ensuring that a user entering a field will be informed of the name of the field), you can use the ```aria-label``` attribute:
+
+```php
+CRUD::field([
+    'name'  => 'email',
+    'label' => 'Email Address',
+    'type'  => 'email',
+    'attributes' => [
+        'aria-label' => 'Email Address',
+    ],
+]);
+```
+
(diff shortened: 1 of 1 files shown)


jorgetwgroup:
Hello @tibbsa thanks for this PR i will ask @pxpm to check about it and let us now if is possible to add this to our documentation.

Thanks a lot again for contribution to our community.

Cheers.

tabacitu:
I absolutely LOVE this. Thank you so much @tibbsa ! Field accessibility is indeed something that we haven't been able to tackle yet - this provides a simple solution to people who need it.

Thank you, merging!

pxpm:
Hey @tibbsa thanks for the PR. 

I don't see a reason why we wouldn't mention it in the docs, very well written I've nothing to say about the PR. 

But seeing that the "output" is the same for all the fields maybe we could ease the developers life by creating a macro `->accessibleLabel(?string $acessibleLabel)` on fields ? 

What would the macro do? 
- Use the string if provided
- Attempt to get the string from label if no string provided
- Resort to using the field name if no string is provided and field has no label
- add to the fields attributes array the `aria-label => $value`

It may be something easy to implement and maybe would help to simplify how accessibility can be achieved and promote it's usage, so that more developers adopt it. 

Ideally it should be "accessible by default", and we won't even need to "enable accessibility". The reason why it's not like that at the moment it's because of the massive change and potentially BC's in lot's of fields. 

Once again, thanks for the PR, I will take this macro idea to my colleagues, but if you like it and are willing to try a PR for it ,go ahead, you can find some examples of field macros we created in the `macros.php` file. 
I can't promise that it will get merged as someone in the team may have a better reason to not merge it than I have to merge, but you can always keep using it in your projects as it seems you will use it alot. 

Cheers

PR #543, opened 2024-01-29
wrong example for daterange

starhobbit:
This comes from client:
'2024-01-21 00:00:00'
'2024-01-29 23:59:59'

we don't need to concat 23:59:59


pxpm:
thanks for the PR @starhobbit 🙏

PR #544, opened 2024-01-30
rephrase custom-views docs

tabacitu:
Heads-up @promatik I felt the need to update the docs because:
- it wasn't obvious to me what the benefit is for the admin;
- we were not showing how it looks (an image is worth 1000 words);

If you need to edit further or I made any mistake, please correct.


tabacitu:
286036856-edb3a77e-4a65-454c-a6dc-eae05837fe3b

karandatwani92:
Hey @tabacitu

People will be able to relate to the given example much easier if the screenshot is in **English**. 

We have mentioned, **"each view has some filters/columns`"** and I just found:
- Filters are not compatible with the `CustomViewOperation`
- `CustomViewOperation` not working on Nested CRUD

I'm about to open an issue on them with details.

PR #547, opened 2024-01-31
Enables time picker for date_range filter

karandatwani92:
Adds guide to:
- Pass options via `date_range_options`
- how to enable time picker for `date_range` filter.


pxpm:
thanks @karandatwani92

PR #563, opened 2024-04-09

peterhauke:

The phrase "This B solution" seems to be in the wrong place or is no longer relevant, so I removed it. However, I am not sure about this.

I rewrote the sentences about why somebody might not like Backpack. I think it is better not to use that kind of negative phrasing in your documentation. The phrasing I have used is more neutral and concentrates of "use" rather than "not like".

### DIFF
--- a/6.x/faq.md
+++ b/6.x/faq.md
@@ -52,9 +52,9 @@ First of all, **run `composer update` on your project**. That will pull in the l
 
 Then you should **re-publish the JS and CSS assets**. There are two ways to do that:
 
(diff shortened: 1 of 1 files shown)


pxpm:
Actually that public publish step is completely removed in Backpack v6. 

Backpack starting v6 uses https://github.com/Laravel-Backpack/basset 

Can you remove the changes done in the publishing step and keep the others so that I can merge them, and I will find time to re-write that bit of docs. 

The asset are in the files that use them as they include the cdn link.

Cheers

peterhauke:
@pxpm I hope that I have removed the changes you mentioned correctly. Thanks.

pxpm:
Perfect @peterhauke 

Really appreciate the effort 🙏 

Cheers

PR #566, opened 2024-04-10

peterhauke:

I removed the semi-colons from the bullet points as most of the bullet points here did not have semi-colons.

I removed the old-fashioned smiley : -) as it looks a little bit odd.

I also noticed that the links to BulkDelete and BulkClone are to the top of the page whereas they could be linked to the particular section.

Also, it appears that the three links at the end for the Custom CRUDs are broken. I think the anchors on the target page have probably changed.

### DIFF
--- a/6.x/getting-started-advanced-features.md
+++ b/6.x/getting-started-advanced-features.md
@@ -2,9 +2,9 @@
 
 ---
 
-**Duration:** 5 min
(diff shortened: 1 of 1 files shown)


pxpm:
Take more emojis @peterhauke 

🥳 🥇 🍾 ❤️ 

Thanks 🙏

PR #567, opened 2024-04-10

peterhauke:

"Mind that" is very colloquial. A simple "however" is much clearer.

I turned FREE and PAID into the standard `span` badges you use and I also changed PAID to PRO.

I think all the times you use "CRUD panel" that "panel" could be all lowercase. I see no reason to capitalize the P here. Also I changed CrudPanel as I think this should also be CRUD panel.

Again, I removed the semi-colons from two sets of bullet points for consistency.

### DIFF
--- a/6.x/getting-started-basics.md
+++ b/6.x/getting-started-basics.md
@@ -4,29 +4,29 @@
 
 **Duration:** 5 minutes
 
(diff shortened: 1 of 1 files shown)


pxpm:
👍 😍 🚀 

Thanks @peterhauke once again! 🙏

PR #568, opened 2024-04-11

peterhauke:

"... we can give you" sounds a little odd.

"use case" is not hyphenated, you use both versions here.


pxpm:
Lovely. ❤️ 

Really appreciate you took all that time to read through the docs and even more that you take time to improve them. 

Thanks @peterhauke 🙏

PR #572, opened 2024-04-25
Fix sample bulkTrash usage where parameter should not be provided.

doncadavona:
The example usage for `bulkTrash` contains the parameter `$id`, which causes errors when copy-pasted from the docs, like:

```
Too few arguments to function App\Http\Controllers\Admin\UserCrudController::bulkTrash(), 0 passed in ...
```

Updating the sample code block to `bulkTrash()` from `bulkTrash($id)` resolved the error.


pxpm:
Thanks @doncadavona 🙏

PR #574, opened 2024-04-29
Added docs for `select2_json_from_api`

promatik:
Related with https://github.com/Laravel-Backpack/PRO/pull/250.

### DIFF
--- a/6.x/crud-fields.md
+++ b/6.x/crud-fields.md
@@ -2349,6 +2349,77 @@ Input preview:
 
 <hr>
 
+### select2_json_from_api  <span class="badge badge-pill badge-info">PRO</span>
+
+Display a select2 that takes its values from an AJAX call.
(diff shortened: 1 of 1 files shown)


6.x/crud-fields.md
```diff
@@ -2349,6 +2349,77 @@ Input preview:
...
+```php
+CRUD::field([
+    'label'                   => 'Airports', // Table column heading
+    'type'                    => 'select2_json_from_api',
+    'name'                    => 'airports', // a unique identifier (usually the method that defines the relationship in your Model)
+    'data_source'             => url('airports/fetch/list'), // url to controller search function (with /{id} should return model)
+
+    // OPTIONAL
+    'delay'                   => 500, // the minimum amount of time between ajax requests when searching in the field
+    'method'                  => 'POST', // route method, either GET or POST
```
pxpm, line 2366:
```suggestion
```

PR #591, opened 2024-08-07
add deploy to production

jcastroa87:
I create a guide with the steps to deploy a Backpack project to Prod.

### DIFF
--- a/6.x/faq.md
+++ b/6.x/faq.md
@@ -114,7 +114,9 @@ If you still can't figure it out, please [open a new discussion in our Community
 <a name="how-do-i-deploy-backpack-to-production"></a>
 ### How do I deploy Backpack to production?
 
-This is a detailed step-by-step guide for deploying a Backpack project to production.
+Deploying a Laravel+Backpack project to production isn't very different from deploying a normal Laravel project to production. You only need to account for Basset, the system in Backpack that publishes the CSS and JS assets your admin panel needs. 
+
+That being said, here's a detailed step-by-step guide to deploying a Backpack project to production, that should work for most production servers: 
 
 1. Local Preparations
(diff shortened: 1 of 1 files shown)


6.x/faq.md
```diff
@@ -110,3 +110,125 @@ If you still can't figure it out, please [open a new discussion in our Community
 - mention the steps you have followed to get there (e.g. `composer require backpack/pro`, `php artisan backpack:require:pro` etc.);
 - include a screenshot of the console output, so we can understand what happened;
 - cross out any personal data (e.g. token username or password); 
+
+<a name="how-do-i-deploy-backpack-to-production"></a>
+### How do I deploy Backpack to production?
+
+This is a detailed step-by-step guide for deploying a Backpack project to production.
```
tabacitu, line 117:
```suggestion
Deploying a Laravel+Backpack project to production isn't very different from deploying a normal Laravel project to production. You only need to account for Basset, the system in Backpack that publishes the CSS and JS assets your admin panel needs. 

That being said, here's a detailed step-by-step guide to deploying a Backpack project to production, that should work for most production servers: 
```

tabacitu:
I love it! Great job @jcastroa87 - I think this is much better than not having a guide at all - we'll have _something_ to point people to, rather than tell them to look for generic tutorials on the internet. I like that it also includes various stuff people might not think of (like optimize). I think some of the things there aren't necessarily needed for Backpack (eg. doing `php artisan schedule:run`), but it's better to have them than not.

Only did one change in the intro - added on paragraph to explain it's not Backpack-specific but Laravel-specific.

@pxpm please give this a read too, and if nothing major comes up, merge and let's run with it 💪

tabacitu:
We will need to add this to v7 docs as well. But we'll do that in the final diff.

PR #592, opened 2024-08-07
Add tutorials links to the docs to make them searchable.

karandatwani92:
- This list will keep growing, so I decided to have a clear, dedicated page for tutorials.
- Here, I also ask the community to submit articles/tutorials. This is the real goal that we want to achieve.


tabacitu:
I like it! The only thing I'd do is move it under Getting Started.

Maybe in the future each tutorial can also have a short description. That way, there are more keywords that can come up in the docs search. But this is already better than nothing so let's go with it!

PR #603, opened 2024-09-26
Fix `exportOnlyColumn` attribute in `crud-operation-list-entries`

aaamra:
change `exportOnlyField` to `exportOnlyColumn`

because in `vendor/backpack/crud/src/resources/views/crud/list.blade.php` the attribute name is `exportOnlyColumn`

and here is the code on your repo 
https://github.com/Laravel-Backpack/CRUD/blob/b58f971aa0f0064f5070ae51f90249d9675c2d50/src/resources/views/crud/list.blade.php#L71-L78


jcastroa87:
Hello @aaamra 

Thanks for this PR, i will ask @pxpm to check and merge.

Thanks again.

Cheers.

PR #619, opened 2025-01-03
add docs for `addClause()` and `addBaseClause()`

tabacitu:
I didn't find any place where we had this documented, so here goes:

### DIFF
--- a/6.x/crud-filters.md
+++ b/6.x/crud-filters.md
@@ -666,3 +666,27 @@ function() { // if the filter is active (the GET parameter "draft" exits)
   // CRUD::addClause('draft');
 });
 ```
+
+
+<a name="when-to-use-addclause-or-add-base-clause-to-apply-filtering-logic"></a>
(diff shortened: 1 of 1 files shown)


pxpm:
Hey @tabacitu thanks for your contribution. 
I 100% LOVE this docs and examples you gave. ❤️ 

We have a little "explanation" about `addBaseClause` here: https://github.com/Laravel-Backpack/docs/blob/c0f76ca1432b5eae731d0683b8bf1578a17fa039/7.x-dev/crud-operation-list-entries.md?plain=1#L247

I think this would be worth to add in both places (here, in filters) and in the list entries docs. Should I copy this over there too ? 

Cheers

tabacitu:
Ah ok, that's great. That's why it didn't show up in search, because it's in a code block. I think it's fine as it is - people can see it in the ListOperation docs, then they'll search for it if they want more info.

We could have provided a link, though 👀

PR #625, opened 2025-03-06
Add basset:clear to upgrade guide

jnoordsij:
When quickly checking out the v7 alpha for possible incompatibilities with my app (first look shows that it's running fine btw, keep up the good work!), I noticed that my local environment erred when trying to render views. Given that this originated from some basset call, I guessed that clearing bassets would solve the issue and it did. I think however that this warrants adding it as a recommended step to the migration guide.


jcastroa87:
Hello @jnoordsij 

Thanks for this PR I will ask @pxpm to check it and if everything is ok we will merge soon.

Cheers.

pxpm:
Hey @jnoordsij hope you are doing well. 

Thanks for the PR 🙏

PR #628, opened 2025-05-30
Docs for the datatable widget

pxpm:
Add docs for the datatable widget

### DIFF
--- a/7.x-dev/base-components.md
+++ b/7.x-dev/base-components.md
@@ -46,6 +46,19 @@ Note that you can further customize this using custom attributes. If you define 
 
 <hr>
 
+<a name="datatable"></a>
+### Datatable
+
+Show a datatable for the defined controller. 
+
+```php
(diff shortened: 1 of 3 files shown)


7.x-dev/base-components.md
```diff
@@ -46,6 +46,34 @@ Note that you can further customize this using custom attributes. If you define
...
+
+The datatable will pick up everything that in your `setupListOperation()`. You can then further add/remove/configure functionality using the configuration closure:
+
+```php
+
+<x-datatable 
+    name="invoices_table"
+    controller="\App\Http\Controllers\InvoiceCrudController" 
+    :configure="function($crud, $parent) { if ($parent) { $crud->addClause('where', 'customer_id', $parent->id); } }" 
+    :updatesUrl="false"
```
tabacitu, line 71: Let's:
- [ ] make this parameter default to `false`;
- [ ] document here what it does (and that it's optional);

PR #631, opened 2025-06-21
add docs for chip widget

tabacitu:
This PR adds docs for the chip widget introduced in https://github.com/Laravel-Backpack/CRUD/pull/5815 and with demo examples in https://github.com/Laravel-Backpack/demo/pull/701

Documented:
- [x] what is a chip
- [x] how to use the `crud::chips.general` chip
- [x] how to use chips as widgets
- [x] how to use chips as columns
- [x] how to use chips outside CrudControllers
- [x] how to create your own chip

Todo:
- [x] add images
- [x] polish text & layout

### DIFF
--- a/7.x-dev/base-widgets.md
+++ b/7.x-dev/base-widgets.md
@@ -349,6 +349,23 @@ class ChartjsPieController extends ChartController
 
 <hr>
 
+<a name="chip"></a>
+### Chip
+
+Shows a chip blade view - which is useful to show more information about a database entry, using little screen real estate.
+
+```php
+[
+    'type' => 'chip',
+    'view' => 'crud::chips.general',
+    'title' => 'invoices',
+    'entry' => Invoice::first(),
+]
+```
+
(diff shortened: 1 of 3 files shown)


tabacitu:
While trying to document this feature, I've realised that... it's difficult to document it, because it's unlike any other features we have:
- it's not a Component, so we can't add it to Components
- it's not a standard CRUD feature, so we can't document it in `fields` or `columns`

Currently I think we have the following options:
- (A) we can turn it into a Component - and maybe call it `Datachip`; but realistically the PHP Class won't do anything; and I do NOT want this to take its info from a controller, that is _overkill_ and would make it more difficult to customize than I intend it to;
- (B) we can keep it a Blade view - and document it in a separate page under the Show & List operations; we basically create a new TYPE of blade file (chips), which comes with one built-in file... and a way to create your own;

From the above... I think I like Option B best. So I'm going to do that.

PR #633, opened 2025-06-30
Update crud-columns.md

flartet:
bad field in join


jcastroa87:
Hi @flartet 

Thanks for this fix.

Cheers.

